Issue #11, opened 2016-10-11
`taylorinteg` doesn't know how to combine mixed input types


PerezHz:
```julia
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.5.1-pre+2 (2016-09-20 03:34 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit f0d40ec* (21 days old release-0.5)
|__/                   |  x86_64-apple-darwin14.5.0

julia> versioninfo()
Julia Version 0.5.1-pre+2
Commit f0d40ec* (2016-09-20 03:34 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i7-4750HQ CPU @ 2.00GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, haswell)

julia> using TaylorIntegration

julia> f(t, x) = x.^2
f (generic function with 1 method)

julia> @time tT, xT = taylorinteg(f, 3, 0, 0.3, 25, 1e-20, maxsteps=50);
ERROR: MethodError: no method matching taylorinteg(::#f, ::Int64, ::Int64, ::Float64, ::Int64, ::Float64; maxsteps=50)
Closest candidates are:
  taylorinteg{T<:Number}(::Any, ::T<:Number, ::T<:Number, ::T<:Number, ::Int64, ::T<:Number; maxsteps) at /Users/Jorge/.julia/v0.5/TaylorIntegration/src/integration_methods.jl:231
  taylorinteg{T<:Number}(::Any, ::Array{T<:Number,1}, ::T<:Number, ::T<:Number, ::Int64, ::T<:Number; maxsteps) at /Users/Jorge/.julia/v0.5/TaylorIntegration/src/integration_methods.jl:272
[...]

PR #13, opened 2016-10-12
Add support for mixed input type variables

PerezHz:
The aim of this PR is to solve #11 ;  namely, to add support for mixed input type variables. This is achieved using `promote`. Consider the following scenario:

```julia
    using TaylorIntegration, ValidatedNumerics
    x0 = @interval 3 #an Interval{Float64}
    t0 = 0 #an Int64
    tmax = @interval 0.333333 #an Interval{Float64}
    abstol = 1e-10 #a Float64

    order = 25
    f(t, x) = x.^2
    tT, xT = taylorinteg(f, x0, t0, tmax, order, abstol, maxsteps=500);
```

Currently (latest master), this will crash, since `taylorinteg` is expecting `x0` (or, for multi-dimensional cases, `eltype(x0)`), `t0`, `tmax`, and `abstol` to be variables of the same type `T`. But, with this PR the above code will run without any trouble. In particular, this PR includes support for validated Taylor integrations, using `ValidatedNumerics`. One can even try "crazy" stuff such as:

```julia
    using TaylorIntegration, ValidatedNumerics
    q0 = @interval 3 # the initial condition as an interval: Interval{Float64}
    t0 = 0 # the initial time: Int64
    x0 = [t0, q0] # the initial condition as an array of intervals: Array{Interval{Float64}}
    tmax = BigFloat(0.333333) # the final time: BigFloat
    abstol = 1e-80 # the absolute (local) tolerance: Float64

    order = 90 # the order of the Taylor expansion: Int64
    f(t, x) = [one(x[1]), x[2]^2]
    tT, xT = taylorinteg(f, x0, t0, tmax, order, abstol, maxsteps=500);
```

And, at the end of the integration, the numerical solution will be a non-trivial, rigorous enclosure of the analytical solution, i.e.:

```julia
exactsol(t, x0) = x0./(1.0-x0.*t) #the analytical solution
exactsol(tT[end], xT[1,2]) ∈ xT[end,2] #note that the first column of xT corresponds to the time, so we have to take the second column of xT
```

will evaluate to `true`, non-trivially! A trivial enclosure would be the interval [-Inf, Inf], but in this case we have a rigorous solution within 10^{-58}:

```julia
julia> radius(xT[end,2])
5.680700231345338420790774078362109301040435921707687232711698988215210593764779e-59
```

Current PR also includes relevant tests for the new functionality.


REQUIRE
```diff
@@ -2,3 +2,4 @@ julia 0.4
 TaylorSeries 0.3.0
 FactCheck 0.4.2
 Compat 0.9.1
+ValidatedNumerics 0.5.0
```
lbenet: `ValidatedNumerics.jl` is used in the notebooks (which does not need require, and in the tests. If it is not used elsewhere, I suggest to create `tests/REQUIRE`, and include this line there.

src/explicitode.jl
```diff
@@ -208,24 +208,45 @@ end
...
+`t_max`, `x(t_max)`), or else when the number of saved steps is larger
 than `maxsteps`.
 
 The integrator uses polynomial expansions on the independent variable
-of order `order` and the parameter `abs_tol` serves to define the
+of order `order`; the parameter `abstol` serves to define the
 time step using the last two Taylor coefficients of the expansions.
 
 The current keyword argument is `maxsteps=500`.
+
```
lbenet: Great idea to include an example of how to use it!

src/explicitode.jl
```diff
@@ -265,13 +286,28 @@ function taylorinteg{T<:Number}(f, x0::T, t0::T, t_max::T,
     return view(tv,1:nsteps), view(xv,1:nsteps)
 end
 
-function taylorinteg{T<:Number}(f, q0::Array{T,1}, t0::T, t_max::T,
-        order::Int, abs_tol::T; maxsteps::Int=500)
+function taylorinteg{S<:Number, T<:Number, U<:Number, V<:Number}(f,
+        q0::Array{S,1}, t0::T, tmax::U, order::Int, abstol::V; maxsteps::Int=500)
+
+    #promote to common type before integrating:
```
lbenet: Lines 294-295, and 298: Isn't it easier to write one single line here? For instance,
```julia
elq0, t0, tmax, abstol = promote(q0[1], t0, tmax, abstol)
```
instead of the three lines?

src/explicitode.jl
```diff
@@ -208,24 +208,45 @@ end
...
+
+`f(t, x) = x^2`
+
+`tv, xv = taylorinteg(f, 3, 0.0, 0.3, 25, 1.0e-20, maxsteps=100 )`
+
+---
 """
-function taylorinteg{T<:Number}(f, x0::T, t0::T, t_max::T,
-        order::Int, abs_tol::T; maxsteps::Int=500)
+function taylorinteg{S<:Number, T<:Number, U<:Number, V<:Number}(f, x0::S,
```
lbenet: I agree that `x0` may be of different type than `t0`, `tmax` or `abstol`; yet, I think we can impose that `t0` and `tmax` are of the same type (the are both times!), and maybe `x0` and `abstol`. What do you think about this? (This comment applies also somewhere below.)

lbenet:
I don't think I mentioned this but, I **really** like the notebook "Validated-Taylor-integration-example.ipynb" !

REQUIRE
PerezHz: I agree! My mistake.. Already removed this file in a new commit

PerezHz: I meant the last comment for `examples/my-first-jt.ipynb`; but I also agree that we should remove ValidatedNumerics from TaylorIntegration's REQUIRE file

PerezHz: Pushed a new commit where I moved the ValidatedNumerics requirement to test/REQUIRE; tests work fine on my machine with this new change, let's see what the opinion of travis is about this

src/explicitode.jl
PerezHz: Well yes, in principle `t0` and `tmax` should be of the same type; but in the following scenarios 

```julia
using TaylorIntegration
@time tT, xT = taylorinteg( (t, x)->x.^2, 3, 0, 0.3, 28, 1e-20 )
@time tT, xT = taylorinteg( (t, x)->x.^2, 3, 0, 1//3, 28, 1e-20 )
@time tT, xT = taylorinteg( (t, x)->x.^2, 3.0, 0, 1/3, 28, 1e-20 )
```

Here, `t0` and `tmax` are variables of different type and my opinion is that `taylorinteg` should be able to handle cases such as these

src/explicitode.jl
PerezHz: Sure, thanks! I have also corrected this, now the code reads:

```julia
...
elq0, t0, tmax, abstol = promote(q0[1], t0, tmax, abstol)
#convert the elements of q0 to the common type:
q0 = map(x->convert(typeof(elq0), x), q0)
...
```

src/explicitode.jl
```diff
@@ -290,12 +290,9 @@ function taylorinteg{S<:Number, T<:Number, U<:Number, V<:Number}(f,
...
     #promote to common type before integrating:
-    promoted_vars = promote(q0[1], t0, tmax, abstol)
-    elq0, t0, tmax, abstol = promoted_vars
-    #W is the common type:
-    W = eltype(promoted_vars)
-    #convert the elements of q0 to the common type W:
-    q0 = map(x->convert(W, x), q0)
+    elq0, t0, tmax, abstol = promote(q0[1], t0, tmax, abstol)
+    #convert the elements of q0 to the common, promoted type:
+    q0 = convert(Array{typeof(elq0)}, q0)
```
lbenet: I think you are aware of this, but I just want to make it explicit: `q0` is actually mutating with respect to type (line 295). Is this intended? If the idea is *not* to change it, then it is perhaps better to use a different variable...

src/explicitode.jl
lbenet: I get your point, and I do not disagree with it.  Yet, `tT` eventually has to be some kind of floating-point representation of a continuous variable, and I guess `xT` too. My suggestion, somehow, was related to this, though it actually does not impose it!

For example, using `taylorinteg( (t, x)->x.^2, 3, 0, 1, 28, 1//10^16 )` yields that `tT` and `xT` are `Rational{Int}` (thanks to `abstol`!), which I wonder if is really what we want to get. Maybe we should impose (in the `promote` line` that the promotion turns whatever into `Float64` at least. I hope to make this clear...

REQUIRE
lbenet: Thanks!

README.md
```diff
@@ -23,6 +23,11 @@ Comments, suggestions, and improvements are welcome and appreciated.
 
 + [x'=x^2](http://nbviewer.jupyter.org/github/PerezHz/TaylorIntegration.jl/blob/master/examples/x-dot-equals-x-squared.ipynb)
 + [Kepler problem](http://nbviewer.jupyter.org/github/PerezHz/TaylorIntegration.jl/blob/master/examples/Kepler-problem.ipynb)
++ [Evaluating definite integrals](http://nbviewer.jupyter.org/github/PerezHz/TaylorIntegration\
+.jl/blob/master/examples/Evaluating-definite-integrals.ipynb)
++ [Damped, driven linear oscillator](http://nbviewer.jupyter.org/github/PerezHz/Ta\
+ylorIntegration\
+.jl/blob/master/examples/Damped-driven-linear-oscillator.ipynb)
```
PerezHz: Sure, I agree! Actually it should be already listed in the examples section of `README.md`

src/explicitode.jl
PerezHz: Well, perhaps in the future we could add support for exotic users who prefer working with `Rational{BigInt}`'s instead of `Float64`'s! But for the time being I agree that all variables should be at least `Float64`'s. So I've changed the promotions in the code for the following:

```julia
# in the line below, x0::T with T<:Number
x0, t0, tmax, abstol, afloat = promote(x0, t0, tmax, abstol, one(Float64))
# and something analogous when x0::Array{T,1}, with T<:Number
```

Is this what you were suggesting?

lbenet: I think it *only* makes sense to impose some sort of `AbstractFloat`'s here. I think what you did is ok.

lbenet:
I think this is ready; I'll merge this.

Thanks!

PR #17, opened 2017-02-06
Integration of ODEs with complex dependent variables

lbenet:
This PR adds the possibility to integrate ODEs with complex dependent variables (the independent variable is still real), and includes some simple tests.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -63,6 +63,7 @@ facts("Tests: dot{x}=x^2, x(0) = 3; nsteps <= maxsteps") do
     @fact length(tv) --> 14
     @fact length(xv[:,1]) --> 14
     @fact length(xv[:,2]) --> 14
+    @fact xv[1,1:end] --> q0
     @fact tv[end] < 1/3 --> true
     @fact tv[end] --> tmax
     @fact xv[end,1] --> xv[end,2]
@@ -88,6 +89,7 @@ facts("Tests: dot{x}=x^2, x(0) = 3; nsteps <= maxsteps") do
     @fact length(tv) --> 28
     @fact length(xv[:,1]) --> 28
     @fact length(xv[:,2]) --> 28
+    @fact xv[1,1:end] --> q0
     @fact tv[end] < 1/3 --> true
     @fact tv[end] --> tmax
     @fact xv[end,1] --> xv[end,2]
@@ -109,6 +111,7 @@ facts("Tests: dot{x}=x.^2, x(0) = [3.0,1.0]") do
 
     tv, xv = taylorinteg(eqs_mov, q0, 0.0, 0.5, _order, _abs_tol)
     @fact length(tv) --> 501
+    @fact xv[1,:] --> q0
     @fact tv[end] < 1/3 --> true
 
     trange = 0.0:1/8:1.0
@@ -116,6 +119,7 @@ facts("Tests: dot{x}=x.^2, x(0) = [3.0,1.0]") do
     @fact size(xv) --> (9,2)
     @fact q0 --> [3.0, 1.0]
     @fact typeof(xv) --> Array{eltype(q0),2}
+    @fact xv[1,1:end] --> q0
     @fact (isnan(xv[4,1]) && isnan(xv[4,2])) --> true
     @fact (isnan(xv[end,1]) && isnan(xv[end,2])) --> true
     @fact abs(xv[3,2] - 4/3) ≤ eps(4/3) --> true
@@ -133,6 +137,7 @@ facts("Test non-autonomous ODE: dot{x}=cos(t)") do
     @fact length(tT) < 501 --> true
     @fact length(xT[:,1]) < 501 --> true
     @fact length(xT[:,2]) < 501 --> true
+    @fact xT[1,1:end] --> x0
     @fact tT[1] == t0 --> true
     @fact xT[1,1] == x0[1] --> true
     @fact xT[1,2] == x0[2] --> true
@@ -144,6 +149,7 @@ facts("Test non-autonomous ODE: dot{x}=cos(t)") do
     @fact length(tT) < 501 --> true
     @fact length(xT[:,1]) < 501 --> true
     @fact length(xT[:,2]) < 501 --> true
+    @fact xT[1,1:end] --> x0
     @fact tT[1] == t0 --> true
     @fact xT[1,1] == x0[1] --> true
     @fact xT[1,2] == x0[2] --> true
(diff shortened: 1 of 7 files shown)


lbenet:
Tests may not pass because some changes have been required in [TaylorSeries.jl](https://github.com/JuliaDiff/TaylorSeries.jl), which are already in master. I need to tag a new version in TaylorSeries in order to merge this, so please, hold me on.

I think tests do pass locally, but please check.

PerezHz:
Thanks for this PR @lbenet ! I did check the tests for this PR locally, and they were successful! Using TaylorSeries latest master, output on my machine was:

```julia
WARNING: deprecated syntax "[a=>b for (a,b) in c]".
Use "Dict(a=>b for (a,b) in c)" instead.

WARNING: deprecated syntax "[a=>b for (a,b) in c]".
Use "Dict(a=>b for (a,b) in c)" instead.
Tests: dot{x}=x^2, x(0) = 1
WARNING: Maximum number of integration steps reached; exiting.
WARNING: Maximum number of integration steps reached; exiting.
11 facts verified.
Tests: dot{x}=x^2, x(0) = 3; nsteps <= maxsteps
42 facts verified.
Tests: dot{x}=x.^2, x(0) = [3.0,1.0]
WARNING: Maximum number of integration steps reached; exiting.
WARNING: Maximum number of integration steps reached; exiting.
14 facts verified.
Test non-autonomous ODE: dot{x}=cos(t)
18 facts verified.
Test integration of ODE with complex dependent variables
23 facts verified.
```

lbenet:
In travis, the tests are not passing for different reasons, essentially because there is one missing method for `evaluate!` in TaylorSeries.jl (for Julia v0.4 and 0.5) and due to depreactions (in Julia v0.6).

At the moment I'm tagging a new version of TaylorSeries.jl, probably the final one supporting Julia 0.4; the idea is to tag a new one (which only supports Julia 0.5 and 0.6) which includes some deprecations for Julia 0.6. This should solve the problems for 0.6 (but creates new for Julia0.4); that new version should be the minimum required for TaylorIntegration.jl, though we would need to drop support for Julia 0.4.

So let's be patient a couple of days...

lbenet:
The current tag for `TaylorSeries.jl` is 0.3.1, and this is the last version supporting Julia v0.4. 

I am about pushing a new commit, where I require 0.3.1 as the minimum version of `TaylorSeries.jl`, and noticed that we REQUIRE `Compat.jl`. It seems to me we do not use it, except for compatibility of `views`, i.e., for Julia v0.4 support. 

My point is that we should think about dropping Julia v0.4 support. Any thoughts @PerezHz ?

PerezHz:
I agree @lbenet! We should focus on supporting Julia 0.5 and 0.6! So yeah perhaps it's best to "un-REQUIRE" Compat.jl and drop support for Julia 0.4.. That also would be consistent with what I understand you're currently doing with TaylorSeries.jl. Just went ahead with this in commit `c4141b0` and removed Compat.jl as well as the conditionals such as:

```julia
if VERSION < v"0.5-"
    @fact xT[1,1:end] --> x0'
else
    @fact xT[1,1:end] --> x0
end
```

which are no longer needed in `src/runtests.jl`

PerezHz:
Incidentally, completed tests so far in travis and appveyor with Julia 0.4 are passing!

test/runtests.jl
```diff
@@ -63,11 +63,6 @@ facts("Tests: dot{x}=x^2, x(0) = 3; nsteps <= maxsteps") do
     @fact length(tv) --> 14
     @fact length(xv[:,1]) --> 14
     @fact length(xv[:,2]) --> 14
-    if VERSION < v"0.5-"
-        @fact xv[1,1:end] --> q0'
-    else
-        @fact xv[1,1:end] --> q0
-    end
```
lbenet: The line `@fact xv[1,1:end] --> q0` should remain there, i.e., where the Julia version is at least 0.5? The same on the other changes below in the file.

lbenet:
Incidentally, travis is taking really too long to complete the tests!

test/runtests.jl
lbenet: Moreover: in `TaylorIntegration.jl` there is the line `import Compat.view` which should also be removed...

PerezHz:
I think the tests on macOS (ie, OS X) are slowing travis...

test/runtests.jl
PerezHz: Sure! My mistake.. I'll correct this right away!

PerezHz: Thanks!!

PerezHz:
@lbenet just checked travis log file, it seems there was a bug in Julia 0.6 which was just fixed a couple hours ago in commit [03278a6](https://github.com/JuliaLang/julia/commit/03278a68f15e45e7f25bba90ca811922d7c79f5f), which essentially repeats endlessly a warning about Arrays. Travis is using an older commit (see TaylorIntegration travis [job #142.2 log file lines 156 and 157](https://travis-ci.org/PerezHz/TaylorIntegration.jl/jobs/200244703) ), and thus the tests run endlessly while displaying Array deprecation warnings.

Still, haven't figured why for [job #140.3](https://travis-ci.org/PerezHz/TaylorIntegration.jl/jobs/200215146) travis used a newer Julia commit, which caused the tests to finish successfully, although with a couple warnings (which will be fixed by TaylorSeries.jl latest release). If we can tell travis to use at least Julia 0.6 commit 03278a6, then tests will finish successfully... in a finite time.

PerezHz:
Restarted the jobs which failed last time on travis, with a newer Julia 0.6 build; tests should pass now

lbenet:
Finally, the tests are passing !

I noticed the problem you mentioned and tried to get rid of such occurrences, though it was done in a different PR in TaylorSeries.jl (see https://github.com/JuliaDiff/TaylorSeries.jl/pull/83), which is more advanced that v0.3.1. I think we should wait until I tag a new version in TaylorSeries.jl (0.4.0), without any support to Julia v0.4, and then include that version in the `REQUIRE` file here. This should happen in the next few days.

lbenet:
Tests in linux (and windows) seem to pass; we should wait to get the tests also in Mac (though they pass locally in mine). Once this happens, this PR can be merged.

I would suggest to solve the deprecation warnings (in another PR) before tagging a new (minor) release. (A minor release, since we will not support Julia 0.4 from here on.) I will try to work on them tonight.

lbenet:
Merging!

PerezHz:
Great!! Thanks!! This is a really cool new feature!

PR #18, opened 2017-02-22
WIP: Improve allocation

lbenet:
This is work in progress. 

I think we could improve the use of memory by allocating in `taylorinteg` the `Taylor1` container, which is updated by the other functions. Like this, we would only have one such allocation for the whole integration, instead of one per step of the integration, which is better at least in terms of `gc` managment.

I have extended this idea to the similar functions in `liapunovspectrum.jl` and `jettransport.jl`. I think the improvement should be better in those latter cases. Yet, I don't have any test case to check nor to measure performance. Tests are passing, but with those I don't see any big difference, nor using some of the notebooks in `examples/`.

@PerezHz Can you see if this idea is somewhat worth to be pursued? Some benchmarks would be excellent!


PerezHz:
Wow, thanks @lbenet! I think this is a great idea! I'll definitely check into this and get back to you with some meaningful benchmarks!

src/explicitode.jl
```diff
@@ -302,10 +302,8 @@ function taylorinteg{T<:Number}(f, x0::T, t0::T, tmax::T, order::Int,
 
     # Integration
     while t0 < tmax
-        # xold = x0
-        # δt, x0 = taylorstep(f, t0, tmax, x0, order, abstol)
         δt, x0 = taylorstep!(f, xT, t0, tmax, x0, order, abstol)
-        xT = Taylor1( x0, order )
+        xT.coeffs[1] = x0
```
lbenet: Indeed, this may save some time, which accumulated may be significant! Yet, I'm not sure if it is not "dangerous", in the sense that for the next call of `taylorstep!` contains a `xT` with the correct constant value, but some *non-zero* coefficients for the rest. While this may be ok, since they are overwritten, it's work putting a comment there!

src/explicitode.jl
```diff
@@ -373,7 +369,7 @@ function taylorinteg{T<:Number}(f, q0::Array{T,1}, t0::T, tmax::T,
         end
     end
 
-    return view(tv,1:nsteps), view(xv,:,1:nsteps)' #for xv, first do view, then transpose (otherwise it crashes)
+    return view(tv,1:nsteps), view(transpose(xv),1:nsteps,:)
```
lbenet: Did you check the effect (on the benchmarks) of this change? I suspect that `transpose` allocates an array, whereas the view does not.

src/jettransport.jl
```diff
@@ -100,60 +100,6 @@ end
 
 # taylorstep and taylorstep!
 doc"""
```
lbenet: Thanks for cleaning my mess !

PerezHz:
My benchmark results and a suggestion: 

For a specific benchmark case I'm using locally (restricted N+1 body problem), there were 172.47 M allocations (14.294 GB) when using latest master. After changing to the lb/alloc_improv branch and running again the same benchmark case, the memory use was slightly better: 172.47 M allocations (14.293 GB).

Then, after reviewing the changes you were suggesting, I thought that perhaps in `taylorinteg` it's better if instead of calling the `Taylor1` constructor, we copy directly the elements of `x0` into `xT`. Testing this idea, I found a further drop on memory use: 172.14 M allocations and 14.262 GB of memory use, corresponding to an improvement of about 0.2%. This latter proposal is contained in commit 274f05c1263e5e2acad52987e2ccf7df5644449c.

Also, I couldn't find a systematic drop in execution times when comparing this PR vs latest master..

lbenet:
It seems that my first comment has no impact in the tests, so just ignore it.

Have you checked the benchmarks?

lbenet:
Your last comment is consistent with what I got. So all this leads to a marginal improvement.

What about including now in place evaluation of the equations of motion?

PerezHz:
Yes, I agree! If we're all happy with the changes made so far, I can work on the in-place evaluation right away!

src/explicitode.jl
PerezHz: Good question! I'll check this more carefully!

lbenet:
I think you simply should do it! 

It is worth thinking where we allocate the vector `xdot`, to avoid re-allocating it.

src/explicitode.jl
PerezHz: Just tried both versions and find no difference (first `transpose` then `view`, and viceversa), I think that upon calling `transpose`, a new array is allocated, and it doesn't matter if we transpose either before of after doing `view`.. So if we can choose, perhaps the code is more understandable writing `view(transpose(xv),1:nsteps,:)`, instead of `view(xv,:,1:nsteps)'`

lbenet: I note the following:
```julia
julia> v = randn(1000, 1000);

julia> b = transpose( view(v, :, :) );

julia> c = view( transpose(v), :, : );

julia> sizeof(v)
8000000

julia> sizeof(b)
8000000

julia> sizeof(c)
24
```

So `c` is truly a view; I thus agree, it seems better to use `view(transpose(xv),1:nsteps,:)`.

lbenet:
I really think we should give a try to in place evaluation of `f` within this PR.

PerezHz:
I'm already working on it! Actually the first changes you made within this PR were actually necessary in order to handle in-place evaluation more easily! I already have a preliminary version which seems to be working locally; I'll work out  all the according changes and upload them for revision!

PerezHz:
This is a preliminary solution for the in-place evaluation of the equations of motion; I did not make any changes to the scalar (ie, 1-dimensional) methods, since it must be handled differently. One option would be to re-interpret the scalar case as a 1 element vector; but I already tested that idea and our current implementation of the scalar case seems to be faster. I also updated all the examples and most docstrings. Test pass locally, but let's see what does travis and appveyor think!

lbenet:
I see that you have updated the example notebooks; is there an improvement with respect to speed and allocation?

PerezHz:
Yes! In the Kepler problem case I found an improvement around 10%, but for the damped driven linear oscillator example and the definite integral example I found an improvement of more than 50%! Those were "quick" benchmarks, I'd prefer to benchmark this PR more systematically before posting "official" performance improvement figures; I'll work on those benchmarks and post my results here.

PerezHz:
Tests are passing!

lbenet:
The subtlety of the 1d case is what made us to have various `taylorintegration` methods. For the 1d case, we could ask that `f(t,u,du)` returns `du` (instead of `nothing`), or pass `du` as a length-1 vector, and only use the value of `du[1]`.

This is a major change, so we should probably write about it in `README.md`, at least to warn the user of the way we require `f`.

PerezHz:
Ok, I agree, I will add those details to `README.md`.. And also will see which option is better for handling the 1d case!

test/REQUIRE
```diff
@@ -1 +1 @@
-ValidatedNumerics 0.5.0
+TaylorSeries 0.4.0
```
lbenet: Good catches!

lbenet: Though, since `using TaylorIntegration` uses `TaylorSeries` (but is not exported) this may not be needed. The only difference is that it may be required to use things as `TaylorSeries.Taylor1`, as it was before. I don't know what's better.

test/runtests.jl
```diff
@@ -198,4 +215,94 @@ facts("Test integration of ODE with complex dependent variables") do
...
 end
 
+facts("Test jet transport: simple pendulum") do
+
+    function pendulum!(t, x, dx) #the simple pendulum ODE
+        dx[1] = x[2]
+        dx[2] = -sin(x[1])
+    end
+
+    const varorder = 2 #the order of the variational expansion
```
lbenet: Do this constants are used elsewhere? Otherwise, maybe it's better to use `local` and enforce the type signature.

test/runtests.jl
```diff
@@ -198,4 +215,94 @@ facts("Test integration of ODE with complex dependent variables") do
...
+        xv_disp = taylorinteg(pendulum!, q0+ξ, tr, _order, _abstol, maxsteps=100)
+        #evaluate jet at q0+ξ
+        xvTN_disp = map( x->evaluate(x, ξ), xvTN )
+        #the propagated exact solution at q0+ξ should be approximately equal to the propagated jet solution evaluated at the same point:
+        @fact isapprox( xvTN_disp, xv_disp ) --> true
+    end
+
+end
+
+facts("Test Lyapunov spectrum integrator: Lorenz system") do
```
lbenet: Good addition!

lbenet:
Tests are passing. Can you address my review observations?

lbenet:
... And perhaps further cleaning up would be ok.

lbenet:
Once you are done, drop me a line; I'm changing the docstrings.

src/liapunovspectrum.jl
```diff
@@ -80,12 +70,11 @@ function classicalGS(A)
     return Q, R
 end
 # Modified Gram–Schmidt (Trefethen algorithm 8.1)
-function modifiedGS(A)
+function modifiedGS!(A, Q, R, aⱼ, qᵢ)
```
PerezHz: Thanks! I'll do the corresponding changes to `classicalGS` as well

src/liapunovspectrum.jl
```diff
@@ -80,12 +70,11 @@ function classicalGS(A)
...
     m,n = size(A)
-    Q = similar(A)
-    R = zeros(eltype(A),n,n)
-    aⱼ = Array{eltype(A)}(m)
-    qᵢ = similar(aⱼ)
+    R[:,:] = zeros(eltype(A),n,n)
```
PerezHz: Before commiting this change I made some tests with `R = ...` instead of `R[:,:] = ...` but it wasn't working so that's why I put the `[:,:]`; using `view` sounds nice, I'll try it!

src/liapunovspectrum.jl
```diff
@@ -111,19 +100,18 @@ function modifiedGS(A)
             Q[ind,j] = vⱼ[ind] / R[j,j]
         end
     end
-    return Q, R
+    #return Q, R
```
PerezHz: ok thanks!

lbenet:
Just to put it clearly: my point with the docstrings is that the info (e.g. `taylorinteg`) is repetitive; the only thing that changes is whether we have `Vector{T}` or `Vector{TaylorN{T}}`. 

I think we should make then the docstrings less repetitve, and perhaps simply include the examples.

PerezHz:
Just pushed a commit before reading your review, I was precisely trying to cleanup a bit our notation, mainly changing from `xdot` to `dx` but also changing from `xT` to `x`, but I also changed the docstrings a bit! Hope our modifications don't overlap! Let me incorporate the changes you mentioned, just give a few minutes

lbenet:
Sure!

test/REQUIRE
PerezHz: I put TaylorSeries there because it simplifies the notation for the jet transport tests

test/runtests.jl
PerezHz: I'll just remove the `const` since it's irrelevant, thanks!

test/runtests.jl
PerezHz: 😄

src/liapunovspectrum.jl
lbenet: Let's simply check it and see what's better.

test/REQUIRE
lbenet: Fine.

test/runtests.jl
lbenet: I'm not sure if this matters or not, but it may be better to return `nothing` from the equations of motion for the multi-dimensional case, and to have the explicit return of `dx` for the one-dim case. If it doesn't matter, it may be a good idea to return `dx` always.

src/liapunovspectrum.jl
PerezHz: Just tried various combinations of `R`, `R[:,:]`, `zeros(eltype(A),n,n)` and `view(zeros(eltype(A),n,n))`, but the only combination that seems to work is the one already implemented

PerezHz:
Ok, just pushed a commit which addresses your review, many thanks for all your comments! I agree with what you're saying about the docstrings by the way..

PerezHz:
linux and windows tests have already passed! They only tests pending in travis are the OS X ones

src/liapunovspectrum.jl
lbenet: Let's just then leave as you have it. We may come back eventually to this. Thanks!

lbenet:
Can I now update the docstrings? I'll send you a proposal before we rebase and merge.

test/runtests.jl
PerezHz: Yes, thanks! I agree that the equations of motion should return `nothing` in the multi-dim case!

src/liapunovspectrum.jl
```diff
@@ -134,8 +122,8 @@ function liap_jetcoeffs!{T<:Number}(eqsdiff, t0::T, x::Vector{Taylor1{T}})
         end
 
         # Equations of motion
-        @inbounds xdot[1:dof] = eqsdiff(t0, xaux[1:dof])
-        stabilitymatrix!( eqsdiff, t0, xaux[1:dof], jjac )
+        @inbounds eqsdiff!(t0, xaux, xdot)
```
PerezHz: Oops! Thanks! I'll just remove it

test/runtests.jl
PerezHz: I agree! I'm adding `nothing` at the end of the multi-dim equations of motion in the tests

PerezHz:
Sorry, I noticed I missed some points of your review, just pushed a last commit which addresses those missed points.. Please, go ahead and update the docstrings!

src/liapunovspectrum.jl
ChrisRackauckas: It's better to do `R .= zero(eltype(A))`. `zeros(eltype(A),n,n)` will allocate an `nxn` before it sets the values.

ChrisRackauckas: or just use `fill!`

PerezHz: Thanks! It does improve performance a bit!

PerezHz:
@lbenet please, have a look at [this TaylorSeries issue](https://github.com/JuliaDiff/TaylorSeries.jl/issues/88) that I just submitted, I think we can further improve `liap_taylorinteg` performance!

lbenet:
@PerezHz Can you take a look on the new docstrings and send feedback? Tests are not passing since we are using functionality of current master of TaylorSeries, but they do pass locally.

PerezHz:
@lbenet many thanks for fixing the docstrings! I take the responsibility for them becoming too repetitive 😅

src/explicitode.jl
```diff
@@ -1,19 +1,23 @@
...
+    jetcoeffs!(eqsdiff, t, x)
 
 Returns an updated `x` using the recursion relation of the
-derivatives from the ODE $\dot{x}=dx/dt=f(t,x)$.
+derivatives obtained from the differential equations
+$\dot{x}=dx/dt=f(t,x)$.
+
+`eqsdiff` is the function defining the RHS of the ODE,
+`x` contains the Taylor1 expansion of the dependent variable(s) and
+`t` is the independent variable. See [`taylointeg`](@ref) for examples
```
PerezHz: there's a missing r in `taylorinteg`

src/explicitode.jl
```diff
@@ -33,20 +37,43 @@ function jetcoeffs!{T<:Number}(eqsdiff, t0::T, x::Taylor1{T})
...
-`t` is the independent variable. `dx` stores an in-place evaluation of
-the equations of motion, whereas `xaux` is an auxiliary variable which helps
-with optimization.
-Initially, `x` contains only the 0-th order Taylor coefficients of
+derivatives obtained from the differential equations
+$\dot{x}=dx/dt=f(t,x)$.
+
+`eqsdiff!` is the function defining the RHS of the ODE,
+`x` contains the Taylor1 expansion of the dependent variables and
+`t` is the independent variable. See [`taylointeg`](@ref) for examples
```
PerezHz: There's an r missing in `taylorinteg`

src/explicitode.jl
```diff
@@ -174,15 +182,21 @@ end
...
-time-step of the integration carried out and the updated value of `x0`.
+One-step Taylor integration for the ODE $\dot{x}=dx/dt=f(t, x)$
+with initial conditions $x(t_0)=x_0$, computed from `t0` up to
+`t1`, returning the time-step of the actual integration carried out
+and the updated value of `x0`.
 
-Here, `x0` is the initial (and returned) dependent variables, `order`
-is the degree used for the `Taylor1` polynomials during the integration
+Here, `f` is the function defining the RHS of the ODE (see
+[`taylointeg`](@ref) for examples and structure of `f`), `x` contains
```
PerezHz: Missing r in `taylorinteg`

lbenet:
Never mind about that. Let's wait that https://github.com/JuliaDiff/TaylorSeries.jl/pull/95 is merged, and most probably tag a new version, and then we merge this.

Incidentally, did you notice that in `jettransport.jl` (I think) I changed a vector `TaylorN{Taylor1{T}}` into a vector `Taylor1{TaylorN{T}}`? I think the latter is correct, since we are integrating `TaylorN{T}` variables.

src/liapunovspectrum.jl
```diff
@@ -1,33 +1,44 @@
 # This file is part of the TaylorIntegration.jl package; MIT licensed
 
 """
-    stabilitymatrix!{T<:Number, S<:Number}(eqsdiff, t0::T, x::Array{T,1},
-        jac::Array{T,2})
-    stabilitymatrix!{T<:Number, S<:Number}(eqsdiff, t0::T, x::Array{Taylor1{T},1},
-        jac::Array{Taylor1{T},2})
+    stabilitymatrix!(eqsdiff!, t0, x, δx, dδx, jac)
```
PerezHz: I agree that `dδx` should be used instead of `δdx`, since `dδx` represents the time-derivative of the variations `δx`

src/liapunovspectrum.jl
```diff
@@ -116,7 +137,7 @@ function liap_jetcoeffs!{T<:Number}(eqsdiff!, t0::T, x::Vector{Taylor1{T}},
 
         # Equations of motion
         eqsdiff!(t0, xaux, dx)
-        stabilitymatrix!( eqsdiff!, t0, xaux[1:dof], δx, δdx, jac )
+        stabilitymatrix!( eqsdiff!, t0, xaux[1:dof], δx, dδx, jac )
         @inbounds dx[dof+1:nx] = jac * reshape( xaux[dof+1:nx], (dof,dof) )
```
PerezHz: I'm thinking that we could further improve allocations if we avoid the use of `reshape`, which allocates a new array, but haven't figured out yet a nice way to do this

src/jettransport.jl
```diff
@@ -113,23 +69,8 @@ function stepsize{T<:Number}(q::Array{Taylor1{TaylorN{T}},1}, epsilon::T)
...
-
-Here, `x0` is the initial (and returned) dependent variables, `order`
-is the degree used for the `Taylor1` polynomials during the integration
-and `abstol` is the absolute tolerance used to determine the time step
-of the integration. If the time step is larger than `t1-t0`, that difference
-is used as the time step.
-"""
-function taylorstep!{T<:Number}(f, x::TaylorN{Taylor1{T}}, t0::T, t1::T,
+# taylorstep!
+function taylorstep!{T<:Number}(f, x::Taylor1{TaylorN{T}}, t0::T, t1::T,
```
PerezHz: Great catch! Hadn't noticed this!

src/explicitode.jl
lbenet: Thanks for noticing this! I've changed the four occurrences!

src/liapunovspectrum.jl
lbenet: 😄

src/liapunovspectrum.jl
lbenet: I guess a `view` is always better. I'll play with this...

PerezHz:
> Let's wait that JuliaDiff/TaylorSeries.jl#95 is merged, and most probably tag a new version, and then we merge this.

Looking forward to that merge in TaylorSeries, so that we can tag TaylorIntegration v0.1.0!

> Incidentally, did you notice that in jettransport.jl (I think) I changed a vector TaylorN{Taylor1{T}} into a vector Taylor1{TaylorN{T}}? I think the latter is correct, since we are integrating TaylorN{T} variables.

I definitely had not noticed! It was a great catch!

lbenet:
Just pushed another commit addressing the typos, and including `@simd` loops for some assignments. I hope this improves at least slightly the performance...

ChrisRackauckas:
>Just pushed another commit addressing the typos, and including @simd loops for some assignments. I hope this improves at least slightly the performance...

SIMD is usually automatically applied by the compiler in v0.5 and beyond. This is because the compiler defaults to `-O3`. So I wouldn't expect this to improve performance. In fact, I would only expect it to change performance if it decreases performance by adding SIMD in a spot that the compiler would have determined as detrimental (or have no effect if the compiler determines it's good to have SIMD there).

lbenet:
Thanks for the comment Chris. Do you suggest to simply take the away? Jorge (@PerezHz) has some cases where he could test performance, so let's give it a try. If they don't do anything I'll certainly take them away.

ChrisRackauckas:
I suggest not using `@simd` unless you really know it should use SIMD. The compiler likely makes better choices than you do though, so I've only had it decrease the speeds of codes (since the compiler realized it would be slower to simd those loops than to not simd them!).

Try it yourself in a bunch of test cases and you'll see what I'm talking about. You will have had to build from source or re-build your system image to see anything though.

lbenet:
I'll take your word! I'll revert that back.

PerezHz:
Commit 5c1f679 passes tests (locally) on Julia 0.5, but it doesn't pass local tests on Julia 0.6. On my machine terminal I ran

```
$ julia0.6 -e 'versioninfo(); include("/Users/Jorge/TaylorIntegration.jl/test/runtests.jl")'
```

And the output was:

```
Welcome to Julia @ Laplace!
Julia Version 0.6.0-pre.alpha.80
Commit c29fc1e (2017-03-08 13:43 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i7-4750HQ CPU @ 2.00GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, haswell)

WARNING: deprecated syntax "abstract Result" at /Users/Jorge/.julia/v0.6/FactCheck/src/FactCheck.jl:46.
Use "abstract type Result end" instead.
Tests: dot{x}=x^2, x(0) = 1
WARNING: Maximum number of integration steps reached; exiting.
WARNING: Maximum number of integration steps reached; exiting.
11 facts verified.
Tests: dot{x}=x^2, x(0) = 3; nsteps <= maxsteps
42 facts verified.
Tests: dot{x}=x.^2, x(0) = [3.0,1.0]
WARNING: Maximum number of integration steps reached; exiting.
WARNING: Maximum number of integration steps reached; exiting.
14 facts verified.
Test non-autonomous ODE: dot{x}=cos(t)
18 facts verified.
Test integration of ODE with complex dependent variables
23 facts verified.
Test jet transport: simple pendulum
14 facts verified.
Test Lyapunov spectrum integrator: Lorenz system
ERROR: LoadError: MethodError: Cannot `convert` an object of type TaylorSeries.Taylor1{Float64} to an object of type Float64
This may have arisen from a call to the constructor Float64(...),
since type constructors fall back to convert methods.
Stacktrace:
 [1] copy!(::IndexLinear, ::Array{Float64,1}, ::IndexLinear, ::Array{TaylorSeries.Taylor1{Float64},1}) at ./abstractarray.jl:641
 [2] convert(::Type{TaylorSeries.Taylor1{Float64}}, ::TaylorSeries.Taylor1{TaylorSeries.Taylor1{Float64}}) at /Users/Jorge/.julia/v0.6/TaylorSeries/src/conversion.jl:10
 [3] copy!(::IndexLinear, ::Array{TaylorSeries.Taylor1{Float64},1}, ::IndexLinear, ::Array{TaylorSeries.Taylor1{TaylorSeries.Taylor1{Float64}},1}) at ./abstractarray.jl:641
 [4] convert(::Type{TaylorSeries.HomogeneousPolynomial{TaylorSeries.Taylor1{Float64}}}, ::TaylorSeries.HomogeneousPolynomial{TaylorSeries.Taylor1{TaylorSeries.Taylor1{Float64}}}) at /Users/Jorge/.julia/v0.6/TaylorSeries/src/conversion.jl:34
 [5] copy!(::IndexLinear, ::Array{TaylorSeries.HomogeneousPolynomial{TaylorSeries.Taylor1{Float64}},1}, ::IndexLinear, ::Array{TaylorSeries.HomogeneousPolynomial{TaylorSeries.Taylor1{TaylorSeries.Taylor1{Float64}}},1}) at ./abstractarray.jl:641
 [6] stabilitymatrix!(::#lorenz!#498{Float64,Float64,Float64}, ::Float64, ::SubArray{TaylorSeries.Taylor1{Float64},1,Array{TaylorSeries.Taylor1{Float64},1},Tuple{UnitRange{Int64}},true}, ::Array{TaylorSeries.TaylorN{TaylorSeries.Taylor1{Float64}},1}, ::Array{TaylorSeries.TaylorN{TaylorSeries.Taylor1{Float64}},1}, ::Array{TaylorSeries.Taylor1{Float64},2}) at /Users/Jorge/TaylorIntegration.jl/test/../src/liapunovspectrum.jl:37
 [7] liap_jetcoeffs!(::#lorenz!#498{Float64,Float64,Float64}, ::Float64, ::Array{TaylorSeries.Taylor1{Float64},1}, ::Array{TaylorSeries.Taylor1{Float64},1}, ::Array{TaylorSeries.Taylor1{Float64},1}, ::Array{TaylorSeries.TaylorN{TaylorSeries.Taylor1{Float64}},1}, ::Array{TaylorSeries.TaylorN{TaylorSeries.Taylor1{Float64}},1}, ::Array{TaylorSeries.Taylor1{Float64},2}) at /Users/Jorge/TaylorIntegration.jl/test/../src/liapunovspectrum.jl:140
 [8] liap_taylorstep!(::Function, ::Array{TaylorSeries.Taylor1{Float64},1}, ::Array{TaylorSeries.Taylor1{Float64},1}, ::Array{TaylorSeries.Taylor1{Float64},1}, ::Array{TaylorSeries.TaylorN{TaylorSeries.Taylor1{Float64}},1}, ::Array{TaylorSeries.TaylorN{TaylorSeries.Taylor1{Float64}},1}, ::Array{TaylorSeries.Taylor1{Float64},2}, ::Float64, ::Float64, ::Array{Float64,1}, ::Int64, ::Float64) at /Users/Jorge/TaylorIntegration.jl/test/../src/liapunovspectrum.jl:166
 [9] #liap_taylorinteg#15(::Int64, ::Function, ::Function, ::Array{Float64,1}, ::Float64, ::Float64, ::Int64, ::Float64) at /Users/Jorge/TaylorIntegration.jl/test/../src/liapunovspectrum.jl:233
 [10] (::TaylorIntegration.#kw##liap_taylorinteg)(::Array{Any,1}, ::TaylorIntegration.#liap_taylorinteg, ::Function, ::Array{Float64,1}, ::Float64, ::Float64, ::Int64, ::Float64) at ./<missing>:0
 [11] (::##482#497)() at /Users/Jorge/TaylorIntegration.jl/test/runtests.jl:300
 [12] facts(::##482#497, ::String) at /Users/Jorge/.julia/v0.6/FactCheck/src/FactCheck.jl:449
 [13] include_from_node1(::String) at ./loading.jl:539
 [14] include(::String) at ./sysimg.jl:14
while loading /Users/Jorge/TaylorIntegration.jl/test/runtests.jl, in expression starting on line 280
```

PerezHz:
BTW, it seems that I have a working "`trange`" version of `liap_taylorinteg`; that is, using `trange::Range{T}` instead of `t0::T`, `tmax::T`

PerezHz:
@lbenet are you having problems with commit 5c1f679 on Julia 0.6 also? Or is it my machine?

lbenet:
I just checked and tests pass locally. I have a bit older version of Julia 0.6 than you: Version 0.6.0-pre.alpha.67 (2017-03-07 03:38 UTC). I'll update Julia and check again...

lbenet:
> BTW, it seems that I have a working "trange" version of `liap_taylorinteg`; that is, using `trange::Range{T}` instead of `t0::T, tmax::T`

Push it up! (Though maybe it should be in another PR.)

lbenet:
I just updated to the current master of Julia 0.6 (Version 0.6.0-pre.alpha.116 (2017-03-10 19:40 UTC)) and the tests pass...

PerezHz:
> I just updated to the current master of Julia 0.6 (Version 0.6.0-pre.alpha.116 (2017-03-10 19:40 UTC)) and the tests pass...

Ok, so I updated to Julia 0.6-pre.alpha.116 (same as yours), and tests weren't passing, and then I realised that I just forgot to checkout TaylorSeries `fuse_operations` on Julia 0.6 😅 now tests are passing! Sorry for making you go through the update 😓

PerezHz:
> Push it up! (Though maybe it should be in another PR.)

I'm for pushing it up, since it's a such nice feature! I'll push it in a couple minutes!

PerezHz:
@lbenet just pushed the new liap_taylorinteg method with corresponding tests.. In the last commit I'm also trying the concept of adding `const` to the arrays we allocate in `liap_taylorinteg` (both methods) to help the compiler a bit, what do you think?

lbenet:
I'm not sure if it helps or not, but it doesn't harm to try!

lbenet:
Just for the record, tests pass locally.

PerezHz:
> Just for the record, tests pass locally.

Great!!

PerezHz:
> I'm not sure if it helps or not, but it doesn't harm to try!

I think that, even if it didn't help performance, adding `const` to allocated arrays inside `taylorinteg` and `liap_taylorinteg` would help with type stability

lbenet:
@PerezHz Due to #22, and the fact that "nighties" refers now to Julia 0.7.0-dev, I suggest to rebase and update appveyor.yml to include Julia 0.6 explicitly. Can you do it? This will trigger running the tests, which is something important now.

PerezHz:
Done as requested, @lbenet 😄! I updated both `appveyor.yml` and `.travis.yml` in order to cover 0.5, 0.6 and 0.7-DEV; and added some more tests

PerezHz:
Merging, since the changes introduced here are dependent on TaylorSeries v0.5.0 (which has already been incorporated into METADATA.jl), and tests are passing

lbenet:
Thanks! I fully agree! Yet, before tagging, change the required version in REQUIRE!

PerezHz:
Sorry, you're right! Had already tagged and submitted to METADATA when I read your comment... But I'm fixing this error right now 😅

lbenet:
I'm sure this @tkelman will notice this.

If you use [attobot](https://github.com/attobot/attobot), you can include that change and re-tag. I learned this the hard way recently :-)

tkelman:
Ooh, good call, I actually did not notice this one, since I don't recognize TaylorSeries' API off the top of my head.

appveyor.yml
```diff
@@ -29,7 +31,7 @@ build_script:
 # Need to convert from shallow to complete for Pkg.clone to work
   - IF EXIST .git\shallow (git fetch --unshallow)
   - C:\projects\julia\bin\julia -e "versioninfo();
-      Pkg.clone(pwd(), \"TaylorIntegration\"); Pkg.build(\"TaylorIntegration\")"
+      Pkg.clone(pwd(), \"TaylorIntegration\"); Pkg.build(\"TaylorIntegration\")
```
tkelman: the trailing quote should still be here, though I guess it worked without?

PerezHz: Thanks! I'll fix this as well! And yeah, appveyor worked without the `"`

PerezHz:
> If you use attobot, you can include that change and re-tag. I learned this the hard way recently :-)

Yes, I'm starting to use attobot! Thanks!

PR #24, opened 2017-06-02
Jp/fixnotebooks

PerezHz:
This PR fixes broken links to examples in `README.md`, as well as adding a new notebook which shows an example of a Lyapunov spectrum computation for the Lorenz system, using TaylorIntegration's `liap_taylorinteg`. Although we added this Lyapunov spectrum integrator since a couple of months, so far we only have examples of Lyapunov spectrum computations at `test/runtests.jl`. What do you think about this, @lbenet? Any feedback would be greatly appreciated!


lbenet:
I restarted the broken test and now is passing. I'm merging. Thanks!

PR #25, opened 2017-06-12
Use Taylor1 for the independent variable in the eqs of motion

lbenet:
We need to use a Taylor1 variable, of appropriate order, when the eqs of
motion are time-dependent. To do so, we pre-allocate a vector, and only
update the zeroth-order value (the actual time).

NOTE: It needs to be checked the Liapunov implementation, since `stability`
matrix was not changed. I think this is ok, since the Jacobian involves only
derivatives with respect to the dependent variables.


lbenet:
Just to make clear the difference with what we had before, the current implementation does not need to introduce another component in the equations of motion, which corresponds to the integration of time.

PerezHz:
Thanks @lbenet for this contribution! It's really nice to add this functionality: it simplifies the notation for ODE functions and also helps with performance for time-dependent problems (at least for the cases I've checked so far)! Tests are passing locally on julia 0.5, 0.6 and 0.7.. I'm working on updating the example notebooks which deal with time-dependent problems!

PerezHz:
Since I didn't change the source code, only the examples, travis and appveyor tests should pass.. I'm really happy with it, so if you don't think there are further changes to be made, feel free to merge as soon as the lights go green!

lbenet:
There are green flags about the tests. I'll merge as you propose.

PR #26, opened 2017-06-14
Suppress use of `.coeffs` where appropriate

PerezHz:
This PR simplifies TaylorIntegration code, with changes introduced in JuliaDiff/TaylorSeries.jl#97. Local tests (julia 0.5.2, 0.6.0-rc3, 0.7-DEV) pass and performance remains the same.

### DIFF
--- a/src/explicitode.jl
+++ b/src/explicitode.jl
@@ -36,7 +36,7 @@ function jetcoeffs!{T<:Number}(eqsdiff, t0::T, x::Taylor1{T}, vT::Vector{T})
         dx = eqsdiff(tT, xaux)
 
         # Recursion relation
-        @inbounds x.coeffs[ordnext] = dx.coeffs[ord]/ord
+        @inbounds x[ordnext] = dx[ord]/ord
     end
     nothing
 end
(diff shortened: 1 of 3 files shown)


PerezHz:
Travis tests are not passing; but upon checking the logs for the failed tests, they *do* pass, but something strange is happening with travis so that it reports the tests as failed... I will wait some minutes and then will restart them

lbenet:
Could the following be relevant: https://discourse.julialang.org/t/julia-v0-7-causing-travis-nightly-build-to-fail/3827/2 ?

PerezHz:
> Could the following be relevant: https://discourse.julialang.org/t/julia-v0-7-causing-travis-nightly-build-to-fail/3827/2 ?

I would be inclined to think that this is a different issue, since we had problems only with Linux with all versions of julia (0.5, 0.6 and 0.7), but travis tests on OSX worked fine on all versions of julia... Tests are finally being reported as passed by travis, and there's only one pending test (build no. 274, job no. 5 on Travis), and I would bet a Californian beer that it will pass! 😄

BTW, what do you think of this PR, @lbenet ? Would you agree to merge these changes, once tests are passing?

PerezHz:
All tests finally passed!

lbenet:
I am in favor of merging this! Thanks a lot!

PR #28, opened 2017-06-14
Use convert(T, Inf) instead of T(Inf) in stepsize

lbenet:
This allows to use the package together with IntervalArithmetics.jl.


PerezHz:
Thanks @lbenet for this contribution! It is really nice to see TaylorIntegration working together with IntervalArithmetics!

Something is funny is happening with PRs at travis (see also #26): tests are actually passing as far as TaylorIntegration is concerned, but for some strange reason, travis is reporting the tests as failing...

Issue #27, opened 2017-06-14
`taylorinteg` not working properly with BigFloats


PerezHz:
On current master, `taylorinteg` breaks down when using BigFloats, if the `maxsteps` keyword argument is greater than the actual number of steps taken during a given integration:

```julia
julia> using TaylorIntegration

julia> function harmosc!(t, x, dx) #the harmonic oscillator ODE
           dx[1] = x[2]
           dx[2] = -x[1]
           nothing
       end
harmosc! (generic function with 1 method)

julia> @time tv, xv = taylorinteg(harmosc!, [0, -1], 0, 20BigFloat(pi), 90, 1.0e-80, maxsteps=15 );
WARNING: Maximum number of integration steps reached; exiting.
  0.663779 seconds (452.26 k allocations: 25.480 MiB, 9.63% gc time)

julia> tv[end] == 20BigFloat(pi)
true

julia> @time tv, xv = taylorinteg(harmosc!, [0, -1], 0, 20BigFloat(pi), 90, 1.0e-80, maxsteps=16 );
ERROR: UndefRefError: access to undefined reference
Stacktrace:
 [1] transpose_f!(::Base.#transpose, ::Array{BigFloat,2}, ::Array{BigFloat,2}) at ./linalg/transpose.jl:54
 [2] transpose(::Array{BigFloat,2}) at ./linalg/transpose.jl:121
 [3] #taylorinteg#4(::Int64, ::Function, ::Function, ::Array{BigFloat,1}, ::BigFloat, ::BigFloat, ::Int64, ::BigFloat) at /Users/Jorge/TaylorIntegration.jl/src/explicitode.jl:453
 [4] (::TaylorIntegration.#kw##taylorinteg)(::Array{Any,1}, ::TaylorIntegration.#taylorinteg, ::Function, ::Array{BigFloat,1}, ::BigFloat, ::BigFloat, ::Int64, ::BigFloat) at ./<missing>:0
 [5] #taylorinteg#3(::Int64, ::Function, ::Function, ::Array{Int64,1}, ::Int64, ::BigFloat, ::Int64, ::Float64) at /Users/Jorge/TaylorIntegration.jl/src/explicitode.jl:408
[...]

PR #29, opened 2017-06-15
Fix some cases where transpose could throw an error

lbenet:
A (non-elegant) solution to #27

### DIFF
--- a/src/explicitode.jl
+++ b/src/explicitode.jl
@@ -449,8 +449,7 @@ function taylorinteg{T<:Number}(f!, q0::Array{T,1}, t0::T, tmax::T,
             break
         end
     end
-
-    return view(tv,1:nsteps), view(transpose(xv),1:nsteps,:)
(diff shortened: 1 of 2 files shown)


PerezHz:
Thanks, @lbenet! While there are still some tests running, so far the tests are all passing! Do you want me to check some performance benchmarks?

lbenet:
@PerezHz There was a strange error in travis (nightly, in linux). Do you have any idea what is the problem?

PerezHz:
> There was a strange error in travis (nightly, in linux). Do you have any idea what is the problem?

@lbenet  As far as the error log shows on both travis and appveyor, somewhere near julia-0.7-DEV.800, something has been broken; do you think we should wait for the error to be solved or should we file a bug report?

PerezHz:
On my machine I'm testing with julia-0.7-DEV.798 and everything is working right!

PerezHz:
Locally, tests are passing on julia 0.5.2, 0.6.0, and 0.7.0-DEV.810

PerezHz:
Merging! Thanks, @lbenet!

PR #31, opened 2017-08-22
Macro wizardry to improve performance

lbenet:
This is work in process towards getting better performance when using `taylorinteg`. I'm just opening this for testing, clarifying cases where the current implementation may not work, for criticism and improvements (among other, about naming of the macro and functions). :-)

The basic usage is through the macro `@taylorize_ode` which should be used in front of the function defining the equations of motion:
```julia
@taylorize_ode function eqdiff!(t, x, dx)
...
end
```

This defines a new method of `TaylorIntegration.jetcoeffs!` (but eventually there will be more), which is so constructed to use less memory, and compute only once each Taylor coefficient of the expansion. The new method may be called (if it exists), which is done internally at `TaylorIntegration.taylorstep!`.

There are still some things to be done, which I am working at at the moment, aside from what it may be discussed here. Among others I mention extend what it has been done to also work with complex dependent variables, or for jettransport. Also, currently the macro does not get generated if the body of the function contains `for` loops, conditionals, and perhaps other possibilities of the language. The possibility for simple tests of the new methods, and documentation clarifying some style of the function which is assumed are also lacking.

Once said all this, I should mention that the tests I have performed show in simple cases  improvements of a factor 4-9 in timing, and something close to ~1/10 in the number of allocations, which scales with the number of equations. So this is really promising. Note that this requires TaylorSeries.jl using what has been coded in the branch "mutating_functions".

Feedback will be appreciated.


PerezHz:
Wow, thanks @lbenet for this contribution!! This work is awesome!! I'll review the code and get back to you!!

PerezHz:
I've working with some examples (Lorenz system, Kepler problem, etc.), and noticed that there was a missing method for `add!` and `subst!` which could help things work smoothly; this issue is addressed in JuliaDiff/TaylorSeries.jl#117

src/parse_eqs.jl
```diff
@@ -0,0 +1,415 @@
...
+"""
+function _rename_indexedvars(fnbody)
+
+    v_indexed = Symbol[]
+    d_indx = Dict{Symbol, Expr}()
+
+    !(isindexed(fnbody)) && return fnbody, v_indexed, d_indx
+
+    # Obtain variables
+    vvars = get_vars(fnbody, rec=true)
```
dfdx: I found myself consistently forgetting to put `rec=true` so I made another function - `find_vars` - which is exactly like `get_vars`, but with `rec=true` by default. 

You may also be interested in `findex()` function which finds an expression matching a pattern. For example, here's how I would write `_rename_indexedvars`:

```
function _rename_indexedvars(fnbody)
    indexed_vars = findex(:(_X[_i...]), fnbody)
    st = Dict(ivar => genname() for ivar in indexed_vars)
    new_fnbody = subs(fnbody, st)
    return new_fnbody, collect(values(st)), Dict(v => k for (k, v) in st)
end
```

lbenet: Thanks for the comment. If you agree, I will incorporate this to the function.

lbenet: Quick question: Does `findex` allows to find cases like ` _x = _x + something`?

dfdx: > Thanks for the comment. If you agree, I will incorporate this to the function.

Sure!

> Quick question: Does findex allows to find cases like _x = _x + something?

Assuming `_x` is matched against the same thing on both sides, yes. All in all, `findex` simply traverses an expression and calls `matchex` to each node, so you can look at `@doc matchex` to learn about other options. 

Since you replace certain parts with the others in your code, you may also be interested in `rewrite()` function which takes an _input pattern_ and _input expression_ and tries to rewrite it according to an _output pattern_. Again, see `@doc rewrite` for details.

lbenet: Thanks for the suggestion, I'll check the docs of both, `matchex` and `rewrite`.

blas-ko:
Can we try to rebase this now that `TaylorIntegration` v0.1.2 is out ? This WIP looks amazing!

blas-ko:
In its current state (latest `mutating_functions` and `parse_eqs` branches) I get:
```julia
 @taylorize_ode function centers_tz!(t,x,dx)
    dx[1] = -x[2]
    dx[2] = +x[1]
    nothing
end

...

taylorinteg(centers_tz!, q0, tv, ord_pol, tol);
  MethodError: no method matching constant_term(::Void)
```
with `q0 = [1.0,0.0]`.

lbenet:
I'll check what's going on, but try to be patient. I have some new commits in this branch, which I haven't pushed because I'm stuck trying to rebase all this. Changes introduced in #36 are causing a head-ache...

blas-ko:
Yes! No hurries! Sorry if I sounded harsh. If there's anything I can do in the meantime please tell me.

lbenet:
I hope to get it done by today...

lbenet:
After some careful sewing I managed to rebase; I just pushed the current version (with `-f`). Be aware that it may not work, but I am on this right now.

PerezHz:
@lbenet would it help if I fix julia 0.6 deprecations on master and tag TaylorIntegration v0.2.0 so that going forward we target in our development julia 0.6+?

PerezHz:
If you agree, then I'd also ammend #32 for a future 0.2.x (or possibly 0.3.0) release

lbenet:
I am using Espresso v0.2.1 (perhaps v0.3.0) which requires Julia 0.6. I just pushed a commit where this is required, after rebasing to current master. Tests pass locally, and I am checking that everything related with parsing the equations is compatible with the changes introduced in #36. Regardless of what travis or appveyor say, I think I will be back to the point I was today.

lbenet:
@blas-ko The problem you encounter is related with the function ending with `nothing` instead of `return nothing`. I just pushed a new commit that should solve this.

lbenet:
... And added tests for `@taylorize_ode` (an awful name!).

PerezHz:
>  And added tests for @taylorize_ode (an awful name!).

Don't have anything in particular against the name `@taylorize_ode`, and don't feel particularly good at naming things, but --just thinking out loud here-- perhaps an alternative name could be `@taylorify` or `@taylorify_ode`?

lbenet:
Rebased again, to current master.

lbenet:
I just pushed two new commits, which include a new test (another implementation of the Kepler problem involving another instance of a `for` loop). It is still not bullet proof, and lots of cases are not yet considered.

Please send me feedback, and ideas to change `@taylorize_ode` into something more attractive.

lbenet:
Travis tests on Julia 0.7 (nighties) are failing, no idea why. Also, appveyor has issues. Any ideas?

PerezHz:
Current problems on travis appear to be related to travis itself, since it fails before it even starts to actually test the code; I've restarted the failing jobs a couple of times to no avail, but I guess this particular problem with travis should be fixed in the coming days... Now, locally the tests are also failing for julia 0.7-DEV, this seems to be related to Espresso:


```julia-repl
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: https://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.7.0-DEV.2157 (2017-10-15 20:04 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit 894b650 (0 days old master)
|__/                   |  x86_64-apple-darwin14.5.0

julia> using Espresso
INFO: Precompiling module Espresso.
WARNING: `isdefined(:symbol)` is deprecated, use `@isdefined symbol` instead
Stacktrace:
 [1] depwarn(::String, ::Symbol) at ./deprecated.jl:68
 [2] include_relative(::Module, ::String) at ./loading.jl:533
 [3] include at ./sysimg.jl:14 [inlined]
 [4] include(::String) at /Users/Jorge/.julia/v0.7/Espresso/src/Espresso.jl:4
 [5] include_relative(::Module, ::String) at ./loading.jl:533
 [6] include at ./sysimg.jl:14 [inlined]
 [7] include(::String) at /Users/Jorge/.julia/v0.7/Espresso/src/Espresso.jl:4
 [8] include_relative(::Module, ::String) at ./loading.jl:533
 [9] include(::Module, ::String) at ./sysimg.jl:14
 [10] anonymous at ./<missing>:2
 [11] eval(::Module, ::Expr) at ./sysimg.jl:53
 [12] process_options(::Base.JLOptions) at ./client.jl:307
 [13] _start() at ./client.jl:391
in expression starting at /Users/Jorge/.julia/v0.7/Espresso/src/utils.jl:4
ERROR: LoadError: LoadError: LoadError: syntax: unexpected "]"
Stacktrace:
 [1] include_relative(::Module, ::String) at ./loading.jl:533
 [2] include at ./sysimg.jl:14 [inlined]
 [3] include(::String) at /Users/Jorge/.julia/v0.7/Espresso/src/Espresso.jl:4
 [4] include_relative(::Module, ::String) at ./loading.jl:533
 [5] include at ./sysimg.jl:14 [inlined]
 [6] include(::String) at /Users/Jorge/.julia/v0.7/Espresso/src/Espresso.jl:4
 [7] include_relative(::Module, ::String) at ./loading.jl:533
 [8] include(::Module, ::String) at ./sysimg.jl:14
 [9] anonymous at ./<missing>:2
in expression starting at /Users/Jorge/.julia/v0.7/Espresso/src/to_einstein.jl:11
in expression starting at /Users/Jorge/.julia/v0.7/Espresso/src/core.jl:28
in expression starting at /Users/Jorge/.julia/v0.7/Espresso/src/Espresso.jl:115
ERROR: Failed to precompile Espresso to /Users/Jorge/.julia/lib/v0.7/Espresso.ji.
Stacktrace:
 [1] compilecache(::String) at ./loading.jl:661
 [2] _require(::Symbol) at ./loading.jl:472
 [3] require(::Symbol) at ./loading.jl:324

julia> 
```

I'm using Espresso's latest master, commit `485449d`.

Now, appveyor errors seem even more strange, the log shows a `GitError`, but I honestly don't know how to intepret that? What I did observe in the log is that the job fails before actually start testing TaylorIntegration, while installing some required packages:

```
INFO: Installing Einsum v0.1.0
INFO: Rolling back install of DataStructures
INFO: Rolling back install of Compat
ERROR: GitError(Code:ERROR, Class:Checkout, Cannot checkout to invalid path 'LICENSE.')
```

Actually, at least on Espresso.jl README file, there's no indication of Windows support. So, until there's Windows support perhaps we shouldn't expect tests to work on appveyor (we could even temporarily allow build failures on appveyor, at least until this is solved)...

PerezHz:
Just noticed that `test/taylorize.jl` hasn't been added to `test/runtests.jl`; therefore, failures can't even be related to TaylorIntegration.jl's tests ran from `test/runtests.jl`... I went ahead and added that line 👍 , but I'd suspect it won't make any difference wrt failing tests

PerezHz:
After the last change I pushed there are some travis errors related to `TaylorSeries._dict_binary_calls` not being defined... I understand that function is defined on the `mutating_functions` branch in TaylorSeries.jl; we could always hack .travis.yml and appveyor.yml with a `Pkg.checkout` to fix this 😉  if that branch isn't ready yet to be merged onto master...

lbenet:
Yes. I haven't merged yet the `mutating_functions` branch of TaylorSeries.jl, which is actually required here. The reason is that I may simplify the approach, or add other functions if needed.

lbenet:
As for this PR, I'd like to have the "ability" to skip functions which are not recognized. What I mean by this is that you may have some external function call that does something specific (within the function defining the ODEs). If that function call is not recognized as a usual funcion (`+`, `sin`, `sqrt`, etc), an error is thrown. I hope to devise a way of avoiding that error and simply copying that function call as it is. So this would provide a partial optimization for those cases.

PerezHz:
Just pushed some travis and appveyor "hacks" (actually, just a couple `Pkg.checkout`s, which should be removed before we merge this PR) which will allow us to get proper testing... Good news is, locally and in a custom branch which I briefly uploaded and then removed, tests are passing in 0.6; but still, I stumbled upon some errors in 0.7:

```
ERROR: LoadError: LoadError: LoadError: LoadError: syntax: expected ")"
Stacktrace:
 [1] include_relative(::Module, ::String) at ./loading.jl:533
 [2] include at ./sysimg.jl:14 [inlined]
 [3] include(::String) at /home/travis/.julia/v0.7/TaylorIntegration/src/TaylorIntegration.jl:3
 [4] include_relative(::Module, ::String) at ./loading.jl:533
 [5] _require(::Symbol) at ./loading.jl:465
 [6] require(::Symbol) at ./loading.jl:324
 [7] include_relative(::Module, ::String) at ./loading.jl:533
 [8] include(::Module, ::String) at ./sysimg.jl:14
 [9] include(::String) at ./sysimg.jl:47
 [10] macro expansion at /home/travis/.julia/v0.7/TaylorIntegration/test/runtests.jl:14 [inlined]
 [11] anonymous at ./<missing>:?
 [12] include_relative(::Module, ::String) at ./loading.jl:533
 [13] include(::Module, ::String) at ./sysimg.jl:14
 [14] process_options(::Base.JLOptions) at ./client.jl:325
 [15] _start() at ./client.jl:391
in expression starting at /home/travis/.julia/v0.7/TaylorIntegration/src/parse_eqs.jl:499
in expression starting at /home/travis/.julia/v0.7/TaylorIntegration/src/TaylorIntegration.jl:13
in expression starting at /home/travis/.julia/v0.7/TaylorIntegration/test/one_ode.jl:3
in expression starting at /home/travis/.julia/v0.7/TaylorIntegration/test/runtests.jl:13
```

If I did everything correctly, then the last travis job logs should reproduce this same error in julia 0.6...

lbenet:
Just pushed a new commit which should address the Julia 0.7 failures.

lbenet:
But Julia 0.7 is still problematic...

blas-ko:
I'm not sure if this is useful at all, but I was looking on how more-than-binary operations are constructed and maybe this can be useful for this PR's purpose. The following code is from julia's [/base/operators.jl](https://github.com/JuliaLang/julia/blob/903644385b91ed8d95e5e3a5716c089dd1f1b08a/base/operators.jl#L424)
```julia
# foldl for argument lists. expand recursively up to a point, then
# switch to a loop. this allows small cases like `a+b+c+d` to be inlined
# efficiently, without a major slowdown for `+(x...)` when `x` is big.
afoldl(op,a) = a
afoldl(op,a,b) = op(a,b)
afoldl(op,a,b,c...) = afoldl(op, op(a,b), c...)
function afoldl(op,a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p,qs...)
    y = op(op(op(op(op(op(op(op(op(op(op(op(op(op(op(a,b),c),d),e),f),g),h),i),j),k),l),m),n),o),p)
    for x in qs; y = op(y,x); end
    y
end

for op in (:+, :*, :&, :|, :xor, :min, :max, :kron)
    @eval begin
        # note: these definitions must not cause a dispatch loop when +(a,b) is
        # not defined, and must only try to call 2-argument definitions, so
        # that defining +(a,b) is sufficient for full functionality.
        ($op)(a, b, c, xs...) = afoldl($op, ($op)(($op)(a,b),c), xs...)
        # a further concern is that it's easy for a type like (Int,Int...)
        # to match many definitions, so we need to keep the number of
        # definitions down to avoid losing type information.
    end
```

Looks like `afold1` does the job. It does recursevely, from first to last, all the operations involved in `op`, each one as a **binary operation**. Note that this only happens when `op` has three arguments or more.

lbenet:
I am not sure if this is useful or not, but I remain open on this. 

Let me try to explain where is the subtle part. As far as I can see, it actually lies in how `TaylorSeries` is implemented. For a concrete example, let's take the addition of three Taylor-whatever quantities, `a+b+c`. Essentially, what we have now is parsed as `res1 = a+b` and `res = res1+c` (or maybe a different combination) which gets translated into `add!(res1, a, b, ord)` and `add!(res, res1,c,ord)`, which mutates `res1` using `a` and `b` at order `ord`, and something similar to `res`. My point is that there is no method `add!(res, a, b, c, ord)`. (Maybe the latter could be a call to `afoldl`, but I do not know that yet.)

The good thing of having all operations as binary or unary within a dictionary is that it is rather straightforward to recognize those defined operations (and parse them), and otherwise copy the (unrecognized) called function; this I do it by checking that the actual call is a key of the relevant dictionary. Currently if the call is not a key of the dictionary it throws an error, but it is not difficult to simply copy the "unrecognized" call. This is not yet implemented/pushed, but I am on this. Once this is done, it will allow to have and outside function "myfun" inside the function defining the ODEs, which should work with Taylor-whatever, to be properly reused (though not optimally) in the new `jetcoeffs!` method.

blas-ko:
> My point is that there is no method add!(res, a, b, c, ord). (Maybe the latter could be a call to afoldl, but I do not know that yet.)

Maybe from the definition of, for example, `add!`, one can implement the methods for `add!(a,b,c,d...)` for Taylors using `afold1` so there's no problem on calling the dictionary at the moment of parsing. (Note that this should be on [this PR]() ) 

This way, `myfunc` calls may be ignored in the way that you say while `op(a,b,c,d...)` can be properly parsed, letting the user be a little more careless in the parenthesis usage for defining `eqs_motion!`.  

Note that I'm just thinking all of this out loud, and I'm not completely sure this is gonna work automatically... but it may be a workaround for this kind of problems.

lbenet:
I agree with you that it is an "uncomfortable feature" for the time being, but it is not related to this PR, but to TaylorSeries.

I shall add the following cases, which simply show that the precedence is introduced in some cases, already at parse time:
```julia
julia> ex = :(z = a+b+c) # no precedence is set
:(z = a + b + c)

julia> ex = :(z = a+b-c) #precedence is set!
:(z = (a + b) - c)

julia> ex = :(z = a*b*c) # no precedence
:(z = a * b * c)

julia> ex = :(z = a/b/c) #with precedence
:(z = (a / b) / c)

julia> ex = :(z = a^b^c) # precedence
:(z = a ^ (b ^ c))
```

So, there are only few cases where it does matter to have parenthesis, and having them helps readability.

PerezHz:
Just pushed a commit which updates travis and appveyor config after merging of JuliaDiff/TaylorSeries.jl#129

lbenet:
Thanks! I am still doing some other tests and changes, so travis, at the moment, is not really a concern :-)

lbenet:
Rebased to current master...

lbenet:
@PerezHz Any idea/suggestion was appveyor is not working at all?

PerezHz:
> @PerezHz Any idea/suggestion was appveyor is not working at all?

I think it's something related to Espresso.jl, or some of its dependencies (Einsum.jl)

PerezHz:
😓 !!! Sorry!!! Just closed this by mistake!!!

PerezHz:
I'll take a look at the appveyor failure more closely

PerezHz:
If I do
```julia
julia> using TaylorIntegration, TaylorSeries

julia> ex = :(function kepler1_parsed!{S<:Number}(t, q::Array{S,1}, dq::Array{S,1})
                  r_p3d2 = Array{S}(1)
                  r_p3d2[1] = (q[1]^2+q[2]^2)^1.5

                  dq[1] = q[3]
                  dq[2] = q[4]
                  dq[3] = mμ*q[1]/r_p3d2
                  dq[4] = mμ*q[2]/r_p3d2

                  nothing
              end)
:(function kepler1_parsed!{S <: Number}(t, q::Array{S, 1}, dq::Array{S, 1}) # REPL[9], line 2:
        r_p3d2 = Array{S}(1) # REPL[9], line 3:
        r_p3d2[1] = (q[1] ^ 2 + q[2] ^ 2) ^ 1.5 # REPL[9], line 5:
        dq[1] = q[3] # REPL[9], line 6:
        dq[2] = q[4] # REPL[9], line 7:
        dq[3] = (mμ * q[1]) / r_p3d2 # REPL[9], line 8:
        dq[4] = (mμ * q[2]) / r_p3d2 # REPL[9], line 10:
        nothing
    end)
```
Then,
```julia
julia> TaylorIntegration._make_parsed_jetcoeffs(ex)
:(function TaylorIntegration.jetcoeffs!(t::Taylor1{T}, q::Vector{Taylor1{S}}, dq::Vector{Taylor1{S}}, ::Type{Val{kepler1_parsed!}}) where {T <: Real, S <: Number}
        order = t.order
        r_p3d2 = Array{S}(1)
        tmp699 = Taylor1(constant_term(q[1]) ^ constant_term(2), order)
        tmp701 = Taylor1(constant_term(q[2]) ^ constant_term(2), order)
        tmp702 = Taylor1(constant_term(tmp699) + constant_term(tmp701), order)
        r_p3d2[1] = Taylor1(constant_term(tmp702) ^ constant_term(1.5), order)
        dq[1] = Taylor1(identity(constant_term(q[3])), order)
        dq[2] = Taylor1(identity(constant_term(q[4])), order)
        tmp705 = Taylor1(constant_term(mμ) * constant_term(q[1]), order)
        dq[3] = Taylor1(constant_term(tmp705) / constant_term(r_p3d2), order)
        tmp707 = Taylor1(constant_term(mμ) * constant_term(q[2]), order)
        dq[4] = Taylor1(constant_term(tmp707) / constant_term(r_p3d2), order)
        for __idx = eachindex(q)
            @inbounds q[__idx].coeffs[2] = dq[__idx].coeffs[1]
        end
        for ord = 1:order - 1
            ordnext = ord + 1
            TaylorSeries.pow!(tmp699, q[1], 2, ord)
            TaylorSeries.pow!(tmp701, q[2], 2, ord)
            TaylorSeries.add!(tmp702, tmp699, tmp701, ord)
            TaylorSeries.pow!(r_p3d2[1], tmp702, 1.5, ord)
            TaylorSeries.identity!(dq[1], q[3], ord)
            TaylorSeries.identity!(dq[2], q[4], ord)
            TaylorSeries.mul!(tmp705, mμ, q[1], ord)
            TaylorSeries.div!(dq[3], tmp705, r_p3d2, ord)
            TaylorSeries.mul!(tmp707, mμ, q[2], ord)
            TaylorSeries.div!(dq[4], tmp707, r_p3d2, ord)
            for __idx = eachindex(q)
                @inbounds q[__idx].coeffs[ordnext + 1] = dq[__idx].coeffs[ordnext] / ordnext
            end
        end
        return nothing
    end)
```

Not the best example, since in this case the `r_p3d2` array declaration is unnecessary, but for other problems it is useful to have auxiliary arrays with intermediate variables which store previously computed expansions. But still, I'd like to point out that in this case, `r_p3d2` should be an `Array{Taylor1{T}}`, but instead is an `Array{T}`. I guess this would be solved if we have the convention that the `q`'s and `dq`'s should be typed as `::Array{Taylor1{S}}` where `S<:Number`? Indeed, if instead I do `ex = :(function kepler1_parsed!{S<:Number}(t, q::Array{Taylor1{S},1}, dq::Array{Taylor1{S},1})`... then everything works fine... So my question is, is this expected behavior?

PerezHz:
Also, `where` syntax seems to be unsupported:

```julia
julia> ex = :(function kepler1_parsed!(t, q::Array{Taylor1{S},1}, dq::Array{Taylor1{S},1}) where {S<:Number}
                  r_p3d2 = Array{Taylor1{S}}(1)
                  r_p3d2[1] = (q[1]^2+q[2]^2)^1.5

                  dq[1] = q[3]
                  dq[2] = q[4]
                  dq[3] = mμ*q[1]/r_p3d2
                  dq[4] = mμ*q[2]/r_p3d2

                  nothing
              end)
:(function kepler1_parsed!(t, q::Array{Taylor1{S}, 1}, dq::Array{Taylor1{S}, 1}) where S <: Number # REPL[6], line 2:
        r_p3d2 = Array{Taylor1{S}}(1) # REPL[6], line 3:
        r_p3d2[1] = (q[1] ^ 2 + q[2] ^ 2) ^ 1.5 # REPL[6], line 5:
        dq[1] = q[3] # REPL[6], line 6:
        dq[2] = q[4] # REPL[6], line 7:
        dq[3] = (mμ * q[1]) / r_p3d2 # REPL[6], line 8:
        dq[4] = (mμ * q[2]) / r_p3d2 # REPL[6], line 10:
        nothing
    end)

julia> TaylorIntegration._make_parsed_jetcoeffs(ex)
ERROR: ArgumentError: It must be a function call:
 function kepler1_parsed!(t, q::Array{Taylor1{S}, 1}, dq::Array{Taylor1{S}, 1}) where S <: Number # REPL[6], line 2:
    r_p3d2 = Array{Taylor1{S}}(1) # REPL[6], line 3:
    r_p3d2[1] = (q[1] ^ 2 + q[2] ^ 2) ^ 1.5 # REPL[6], line 5:
    dq[1] = q[3] # REPL[6], line 6:
    dq[2] = q[4] # REPL[6], line 7:
    dq[3] = (mμ * q[1]) / r_p3d2 # REPL[6], line 8:
    dq[4] = (mμ * q[2]) / r_p3d2 # REPL[6], line 10:
    nothing
end
Stacktrace:
 [1] _extract_parts(::Expr) at /Users/Jorge/TaylorIntegration.jl/src/parse_eqs.jl:100
 [2] _make_parsed_jetcoeffs(::Expr, ::Bool) at /Users/Jorge/TaylorIntegration.jl/src/parse_eqs.jl:47 (repeats 2 times)
```

lbenet:
Thanks for reporting. 

Using the macro *assumes* that the parsed function is coded generically, something like
```julia
function kepler1_parsed!(t, q, dq)
    ...
end
```
Note that I am not declaring the type of `t`, `q` and `dq`. In this case, `where` is not necessary.

Why do you need `r_p3d2` to be a vector of length 1?

lbenet:
Once https://github.com/JuliaDiff/TaylorSeries.jl/pull/162 is merged (and tagged), tests should pass in Julia 0.6. To check this, use the branch "typestable_mutfuncts" of TaylorSeries.

ChrisRackauckas:
https://github.com/PerezHz/TaylorIntegration.jl/pull/50 might reduce the type information to instead only have the function instead of the whole problem enclosed.

lbenet:
I'm checking this right now! Thanks!

lbenet:
I just moved the commits of this PR after #50, and run the tests. A similar error happens:
```julia
[...]
Test integration of ODE with numbers in common interface: Error During Test
  Got an exception of type TypeError outside of a @test
  TypeError: Type: in parameter, expected Type, got TaylorIntegration.##10#14
  Stacktrace:
   [1] taylorstep!(::Function, ::TaylorSeries.Taylor1{Float64}, ::Array{TaylorSeries.Taylor1{Float64},1}, ::Array{TaylorSeries.Taylor1{Float64},1}, ::Array{TaylorSeries.Taylor1{Float64},1}, ::Float64, ::Float64, ::Array{Float64,1}, ::Int64, ::Float64) at /Users/benet/.julia/v0.6/TaylorIntegration/src/explicitode.jl:188
[...]
```

ChrisRackauckas:
Is TaylorIntegration.jl only compatible with named functions? What happens if you use a closure or an anonymous function directly? What exactly are you using the type parameter for?

lbenet:
It does work with anonymous functions.
```julia
julia> using TaylorIntegration

julia> taylorinteg((t,x)->x, 0.5, 0.0, 1.0, 50, 1e-20)
([0.0, 1.0], [0.5, 1.35914])
```

lbenet:
The macro in which I've been working does require named functions (and does not allow dot operations).

lbenet:
> What exactly are you using the type parameter for?

The short answer is that `applicable` uses it through `Val{f}`. I use `applicable` to see if such a method exists, and if so, use it; otherwise it runs the default one (which requires `f`).

The long answer is that the macro creates a specialized method of `jetcoeffs!` which runs much faster. That method has a different signature for dispatch, one of the form `jetcoeffs!(::Taylor1{T}, ::Vector{Taylor1{T}}, ::Vector{Taylor1{T}}, ::Type{Val{f}})`, so it is not only a new method, but a specialized one in the function `f`.

ChrisRackauckas:
>so it is not only a new method, but a specialized one in the function f.

Why not just `jetcoeffs!(::Taylor1{T}, ::Vector{Taylor1{T}}, ::Vector{Taylor1{T}}, ::F) where F` and pass the function instead of its type? Otherwise the value type could have issue with large nested types?

lbenet:
I'll test that; thanks!

lbenet:
The problem remains...

lbenet:
I just pushed two new commits; tests pass locally. The problem with the tests in `jettransport.jl` was solved simply be permuting the first two entries; I've also noticed certain sensitivity in the `bigfloats.jl` tests. 

I don't quite know, but it seems that `Val{f}` used in applicable needs that the function signature is not too long, and parameterized functions are then uncomfortable.

lbenet:
@PerezHz Can you check if the tests pass or not in your machine?

PerezHz:
> @PerezHz Can you check if the tests pass or not in your machine?

Tests do pass locally!! 👏 👏 👏

lbenet:
Restarting travis tests...

lbenet:
Tests in travis are still failing...

lbenet:
I changed slightly travis, and now tests are passing!. Yet, appveyor behaves differently. Can you suggest something on this, @PerezHz ?

PerezHz:
Sorry for the late reply... Seems that appveyor failure is actually related to Einsum.jl, or at least that's what I'd guess from the log, since after trying to install Einsum, appveyor "rolls back" installations of Einsum, Compat and DiffEqBase, and then reports failure, without even starting to actually test TaylorIntegration code

PerezHz:
I just reproduced the issue on a fresh Julia package [PerezHz/MyPkg.jl](https://github.com/PerezHz/MyPkg.jl) with files automatically generated on julia 0.6. REQUIRE on this package is identical to TaylorIntegration's; and `test/runtests.jl` only tests for `@test 1==1` and `@test 2==2`:

screen shot 2018-04-23 at 18 17 20

On the other hand, travis is also activated for this testing package, and tests pass there; so the problem does seem to be related to Einsum.jl on windows...

PerezHz:
So I changed the REQUIRE file on PerezHz/MyPkg.jl to the following two lines:

```
julia 0.6
Einsum 0.1.0
```

And the problem persists:

screen shot 2018-04-23 at 18 51 27

lbenet:
See https://github.com/dfdx/Espresso.jl/issues/38

lbenet:
It seems to me that the problem is related to using `sanitize` [here](https://github.com/PerezHz/TaylorIntegration.jl/blob/db5abd859aace64f575b0a86f4d3b07407cdd54c/src/parse_eqs.jl#L697); tests pass locally without it. I'll check this carefully later and push a commit to amend this.

PerezHz:
> It seems to me that the problem is related to using sanitize here; tests pass locally without it. I'll check this carefully later and push a commit to amend this.

Thanks for checking this! After latest changes in Espresso's master, I pushed a commit here which adds a `Pkg.checkout("Espresso")` in appveyor.yml; tests in appveyor should pass

PerezHz:
Forgot to do the same for travis... will do it right away!

PerezHz:
@lbenet So, after getting the green lights, should I revert the commits I pushed and `git push -f` to this branch right away? Or should we wait for another minor tag release from Espresso? What do you suggest?

lbenet:
Appveyor is already green, and from the tests we did locally, I think travis will also have the green lights. 

I think we should wait to have a new tag of Espresso, so everything is more stable and reproducible. i would also like to increase a bit the coverage, which is now strongly reduced.

It's clear we are close to merge finally this!

PerezHz:
> I think we should wait to have a new tag of Espresso, so everything is more stable and reproducible. i would also like to increase a bit the coverage, which is now strongly reduced.

Ok, cool! Sure, I agree we should increase coverage!

> It's clear we are close to merge finally this!

👏👏👏!!!!

dfdx:
I've created a PR for Espresso v0.4.1: JuliaLang/METADATA.jl#14386

PerezHz:
Just added a couple of tests in an effort to improve coverage 😄 

I think an important TODO is to cover the case of the Lyapunov spectrum integrator; I'll try to give it a shot and see if we can include it here; otherwise, we could include it in another PR

lbenet:
> I think an important TODO is to cover the case of the Lyapunov spectrum integrator

I completely agree that this should also be implemented. Yet, I think that should be left to another PR.

PerezHz:
> I completely agree that this should also be implemented. Yet, I think that should be left to another PR.

OK, agreed!

Just updated REQUIRE for Espresso v0.4.2, which has already been merged to METADATA.jl, and updated accordingly appveyor and travis 👍 Now I'll try to push some more tests to improve coverage

PerezHz:
> I've created a PR for Espresso v0.4.2: JuliaLang/METADATA.jl#14386

Many thanks, @dfdx !

lbenet:
I rebased to current master, and I'm solving issues related to that as well as upgrading things to Julia 0.7 and 1.0.

lbenet:
Tests passing locally. I needed to use `try-catch` block, instead of checking with `applicable` for the existence of the method, because the common interface with DiffEqs test (`test/common.jl`) were throwing an error.

For the time being, this is ok.

lbenet:
@PerezHz Thanks! Tests are passing now also on Julia 1.0 and nighties!

I am working on increasing the coverage of the tests. Also, I'd like to wait until #55 is merged, since I think the macro, or an alternative, could help to speed up further the calculation of the spectrum.

PerezHz:
@lbenet now that #55 is merged, should we rebase this PR to current master?

lbenet:
Rebased to current master...

lbenet:
Thanks for the additions! I just updated my local branch.

I am not quite sure I understand your point; I guess you mean on the `jetcoeffs!` call. Is this correct? In any case, the calculation of the spectrum is now taylorized by hand, which is ok. 

Let's try to get #56 merged, and then rebase this.

PerezHz:
> I am not quite sure I understand your point; I guess you mean on the jetcoeffs! call. Is this correct? In any case, the calculation of the spectrum is now taylorized by hand, which is ok.

Exactly, I mean to use the `try` `catch` block for the `lyap_jetcoeffs!` call in `lyap_taylorstep!` as is done in this PR in `taylorstep!`.

> Let's try to get #56 merged, and then rebase this.

Sure!

PerezHz:
Thanks for rebasing!

lbenet:
Rebased to current master...

lbenet:
(Sorry for the delay... problems with the connection)

lbenet:
I pushed a new commit with the `try-catch` block as discussed above.

lbenet:
@PerezHz Can you perform some benchmarks using the last commit to calculate the Lyapunov spectrum?

PerezHz:
> @PerezHz Can you perform some benchmarks using the last commit to calculate the Lyapunov spectrum?

Sure! On my machine I ran the tests in `test/lyapunov.jl`, with julia 1.0 and TaylorIntegration latest master:
screen shot 2018-09-28 at 21 04 54
Then, I did the same checking out `parse_eqs` and adding a `@taylorize` to the definition of `lorenz!` in `test/lyapunov.jl`. I got the following:
screen shot 2018-09-28 at 21 05 13

PerezHz:
So, yes, there's a considerable speedup which is a lot more noticeable in the case when the user provides `jacobianfunc!`

lbenet:
I think it is some 20-30% improvement in time, though the number of allocations is more drastically reduced.

lbenet:
Rebased to current master; this will allow to write the documentation to use `@taylorize`

lbenet:
@PerezHz I just pushed another bunch of commits; the last one, especially, adds further performance improvements. I've left some `@time` in `test/taylorize.jl` so you can check the timings before `@taylorize` and after it.

Comments are welcome!

lbenet:

src/explicitode.jl
```diff
@@ -84,6 +82,21 @@ function jetcoeffs!(eqsdiff!, t::Taylor1{T}, x::AbstractVector{Taylor1{U}},
...
+
+# __jetcoeffs
+"""
+    __jetcoeffs!(::Val{bool}, f, t, x)
+    __jetcoeffs!(::Val{bool}, f, t, x, dx, xaux)
+
+Chooses a method of [`jetcoeffs!`](@ref) (hard-coded or generated by
+[`@taylorize`](@ref)) depending on `bool::Bool`.
+"""
+@inline __jetcoeffs!(::Val{false}, f, t, x) = jetcoeffs!(f, t, x)
```
PerezHz: @lbenet this is really neat! Just great that you figured out how to select the different cases actually in a Julian way of doing things!

PerezHz:
@lbenet many thanks for the latest changes you introduced, and especially for the nice, Julian solution to the case-selection headache we were having! Just pushed some commits with some minor changes to docs and tests.

lbenet:
Rebasing to current master.

lbenet:
I think it is time to consider merging this. 

It certainly needs more comments in `src/parse_eqs.jl` (which is something I am doing at the moment), cleanup the tests (which is in my to-do list) and improve the docs. Yet, only by exposing it we may get independent feedback to improve it. @PerezHz, can you take a look on this?

PerezHz:
Just pushed a commit which fixes a teeny-tiny text typo in the tests

> It certainly needs more comments in src/parse_eqs.jl (which is something I am doing at the moment), cleanup the tests (which is in my to-do list) and improve the docs. Yet, only by exposing it we may get independent feedback to improve it.

As far as functionality is concerned, I agree that we should merge this PR and should any issues arise then solve those accordingly. Test-wise, I think we can add some more tests cases; I could add some more of those right away, if needed.

lbenet:
Please add whatever tests you think are worth including. I'm adding some comments to all the code, trying to make it easier to follow, and to help not forgetting everything. 

Ping me once you pushed the new tests.

lbenet:
Thanks for adding the tests for jet transport; really neat!

I just pushed a new commit adding comments in `parse_eqs.jl` and few precisions in the documentation. We should wait for the green lights of travis, and then merge. 
Do you agree @PerezHz ?

PerezHz:
> Thanks for adding the tests for jet transport; really neat!

No problem! 😄 

> I just pushed a new commit adding comments in parse_eqs.jl and few precisions in the documentation. We should wait for the green lights of travis, and then merge.
Do you agree @PerezHz ?

I agree, just let me push a last commit with more jet transport tests

PerezHz:
Done! 😄

PerezHz:
There seems to be a really odd issue with the jet transport tests, which is showing in both appveyor and travis on julia 0.7.0; I'll check to see if I can reproduce it locally

PerezHz:
I was able to reproduce locally and pushed a commit which solves the problem (locally, at least). Let's hope travis and appveyor give us now the green lights!

PerezHz:
Green lights at last! Should this be squashed and merged?

lbenet:
I'll go ahead and merge this. Thanks for the great help!

PR #33, opened 2017-08-22
Split tests and use Base.Test

lbenet:


lbenet:
I think this is ready... @PerezHz , can you take a look on this?

PerezHz:
This is looking good for 0.5 and 0.6, but for 0.7-DEV I haven't been able to get tests to pass on my machine... I noted that nightlies are also not passing on neither travis nor appveyor... I still haven't figured the exact reason, but things seem to be pretty weird at the moment: on `test/runtests.jl`, if I comment `complex.jl` from the test file list, then tests pass; but if I comment `one_ode.jl` and `many_ode.jl`, but not `complex.jl`, then tests also pass... So I would think that the problem is not related to the testing code itself but rather it seems that it's more julia-related than TaylorIntegration-related. I'm working with julia 0.7.0-DEV.1454... Do tests pass on your machine with julia 0.7-DEV?

lbenet:
I didn't cheeck the specific cases as you did. Just for the record, the weird message in travis is `AssertionError(msg="invalid age range update")`.

lbenet:
See https://github.com/JuliaLang/julia/issues/21653

PerezHz:
> See JuliaLang/julia#21653

Thanks!! Glad to know we're not alone dealing with this issue!

Update: just updated my julia 0.7 installation to 0.7.0-DEV.1463, and the problem persists...

lbenet:
@PerezHz Thanks for making the tests pass in 0.7! Do you understand why the coverage has decreased so much?

PerezHz:
> @PerezHz Thanks for making the tests pass in 0.7!

Sure! For the record, actually we fixed the problem by declaring the functions in `test/complex.jl`, `test/jettransport.jl` and `test/lyapunov.jl` outside the `@testset` macro.

> Do you understand why the coverage has decreased so much?

No, good catch, I'll look into it!

PerezHz:
> Do you understand why the coverage has decreased so much?

It seems to be related to the fact that, for a strange reason, coveralls thinks we are not testing the Lyapunov spectrum functions contained in `src/liapunovspectrum.jl`, although we are testing them in `test/lyapunov.jl`

PerezHz:
But that report is from travis job 335, we will have to wait until travis finishes job 341 (currently running) to see if this is fixed by then

PerezHz:
After that coverage mishap is fixed, I think this is ready to be merged!

PerezHz:
I added more jet transport tests, this should further improve testing coverage

PerezHz:
There seems to be something broken with codecov, hence the weird coverage %... Still, coveralls is reporting 95% coverage, and all tests are passing, so I'll go ahead and merge this

lbenet:
Thanks!

PR #35, opened 2017-08-27
Simplify code for initial conditions whose elements are parametrised types

PerezHz:
This PR is an implementation of a suggestion made by @lbenet aimed at simplifying TaylorIntegration's code for initial conditions consisting of elements of parametrised types, e.g., `Complex{T}`, `Taylor1{T}`, `TaylorN{T}`, etc...


lbenet:
Sorry for the noise, it seems that TaylorSeries cares about things, so such case can actually not happen.

I'm merging. Thanks!

Issue #37, opened 2017-09-14
Backwards integration


blas-ko:
Right now, it's not possible to integrate ODEs backwards. I know it's that way by design, but why is it that way? 
Sometimes it's practical to integrate "to the past" and I think it would be a nice feature to have. 
I found this by trying to do a Newton's method around an specific time.

PerezHz:
Sure, that's a case that actually hasn't been implemented, and actually there's no reason *why* we haven't done it! But at the same time it shouldn't be much of a trouble to do: if we want to integrate from `t0` to `tmax` with `t0>tmax` I guess a lazy way to do it would be to define, outside `taylorinteg`, `u0=-t0`, `umax=-tmax`, and then integrate from the initial time `u0` through the final time `umax`. As always, we are happy to receive contributions, so if you have some concrete ideas in mind to implement this, please go ahead and submit a PR!

By the way, I don't know if the following is related, but it sounds like it is: there's a trick which @lbenet taught me related to root-finding with Newton's method while performing Taylor integration for a given problem. The essential idea is that, for a given surface `g(t, x, dx)` (`x` is the dependent variable, either a scalar or a vector, and `dx` is the time-derivative of `x`) for which you'd like to test surface crossings, you don't actually need to "re-integrate" the solution while getting the iterates for Newton's method, but a single Taylor expansion near the crossing is sufficient. Let's say that you found there's a crossing through the surface `g` between times `t0` and `t1`, and that you have the Taylor expansion `x(t1)` of the dependent variable `x` at time `t1`. Then, using this expansion, you can evaluate `g1 = g(t, x, dx)(t1)` as well as the time-derivative of g at `t1`, `dg1=derivative(g(t, x, dx))(t1)`. Using these values you can actually compute the first iterate of your Newton process `tnr=t1-g1/dg1`, and iterating this process, you can actually find "a" root of `g` between `t0` and `t1` (there are some technical caveats regarding the unicity of that root within `t0` and `t1`, but that's another story). This way, you avoid the need to integrate, sometimes backwards, sometimes forward, while iterating the Newton process.

Still, backwards integration is a nice feature to have! So, again, please go ahead!

[...]

PR #38, opened 2017-09-16
Add root-finding methods

PerezHz:
I'm opening this PR, as suggested by @blas-ko, to discuss content in the `jp/surface-intersection` branch.

This PR adds root-finding (aka surface-crossings, event location, event detection) via method overloading for `taylorinteg`. For a vectorial differential equation `f!(t, x, dx)`, initial condition `q0`, initial time `t0`, final time `tmax` and a scalar event function `g(t,x,dx)` (whose roots represent the events of interest for detection during the integration of `f!`), the `taylorinteg` call is the following:

```julia
tv, xv, tvS, xvS, gvS = taylorinteg{T<:Real, U<:Number}(f!, g, q0::Array{U,1}, t0::T, tmax::T,
        order::Int, abstol::T; maxsteps::Int=500, nriter::Int=5, eventorder::Int=0)
```

where `order`, `abstol` and `maxsteps` have the same meaning for other  `taylorinteg` calls. The extra keyword arguments `nriter` and `eventorder` represent, respectively, the number of iterations for the Newton-Raphson process which computes the roots of `g`, and `eventorder` represents the order of the derivative of `g` whose roots are of interest. That is, if `eventorder==0`, then `taylorinteg` finds the doots of `g`; if `eventorder==1`, `taylorinteg` finds the root of the 1st-derivative of `g`, etc... The derivatives of `g` are automatically computed with AD techniques, with the help of TaylorSeries.jl.

For general guidelines regarding the implementation, see [this comment](https://github.com/PerezHz/TaylorIntegration.jl/issues/37#issuecomment-329337238) in issue #37.

EDIT: if called this way, `taylorinteg`, besides returning `tv` and `xv` as usual, returns `tvS` (the time for the occurence of the event), `xvS` (the value of the dependent variables at the time of the event) and `gvS` (the function `g` --or its `orderevent`th-derivative-- at the time of the event).


PerezHz:
NOTE: Tests are failing, since they depend upon JuliaDiff/TaylorSeries.jl#118 being merged. If one wants to perform local tests, then the `jp/functors` branch on the fork PerezHz/TaylorSeries.jl should be checked out.

Also, I would suggest that the [`deriv`](https://github.com/PerezHz/TaylorIntegration.jl/pull/38/files#diff-bc0a418a584691d270592db7f74b4330R5) function added here should instead substitute the [`derivative(n, a)`](https://github.com/JuliaDiff/TaylorSeries.jl/blob/347fce177faf30535409fbf9c8b9d04492a6ab5b/src/calculus.jl#L30) method in TaylorSeries; what do you think about this @lbenet?

src/surfaceintersection.jl
```diff
@@ -0,0 +1,111 @@
...
+
+    # Integration
+    nsteps = 1
+    nevents = 1 #number of detected events
+    nextevord = eventorder+1
+    while t0 < tmax
+        δt_old = δt
+        δt = taylorstep!(f!, t, x, dx, xaux, t0, tmax, x0, order, abstol)
+        g_val = g(t, x, dx)
+        if surfacecrossing(g_val_old, g_val, eventorder)
```
blas-ko: I understand this is intended as a _Poincaré Mapping_ detector. In this sense, it's not enough to just detect when does the solution `x(t)` crosses the surface `g`, but also it's important that the crossing is done in the same direction as `q0`. 
```julia
parallelcrossing{T<:Real}(x_last::Vector{T},x_old::Vector{T},q0::Vector{T}) = (x_last - x_old)⋅q0 > zero(T)
parallelcrossing{T<:Real}(x_last::Vector{Taylor1{T}},x_old::Vector{Taylor1{T}},q0::Vector{Taylor1{T}}) = ((x_last - x_old)⋅q0)[1] > zero(T)
parallelcrossing{T<:Real}(x_last::Vector{TaylorN{T}},x_old::Vector{TaylorN{T}},q0::Vector{TaylorN{T}}) = ((x_last - x_old)⋅q0)[1][1] > zero(T) 

poincarecrossing(g,g_old,order,x_last,x_old,q0) = surfacecrossing(g,g_old,order) && parallelcrossing(x_last,x_old,q0)
```
would do the job. Anyways, I don't know if it's intended for Poincaré Mappings or just for surface crossings, so we can discuss about this suggested change.

src/surfaceintersection.jl
```diff
@@ -0,0 +1,111 @@
...
+            x_dx[1:dof] = x
+            x_dx[dof+1:2dof] = dx
+            g_dg[1] = deriv(eventorder, g_val)
+            g_dg[2] = derivative(g_dg[1])
+
+            #Newton-Raphson iterations
+            dt_nr = dt_li
+            evaluate!(g_dg, dt_nr, view(g_dg_val,:))
+
+            for i in eachindex(nrinds)
```
blas-ko: Maybe here we can think of a _convergence indicator_ and take `nrinds` or `maxnrinds` as the maximum number of iterations. We can't assure in this way that the `surfacecrossing` is sufficiently near, or that newton-rapson is converging at all.

src/surfaceintersection.jl
```diff
@@ -0,0 +1,111 @@
+surfacecrossing{T<:Real}(g::Taylor1{T}, g_old::Taylor1{T}, order::Int) = g[order+1]*g_old[order+1] < zero(T)
+surfacecrossing{T<:Real}(g::Taylor1{Taylor1{T}}, g_old::Taylor1{Taylor1{T}}, order::Int) = g[order+1][1]*g_old[order+1][1] < zero(T)
+surfacecrossing{T<:Real}(g::Taylor1{TaylorN{T}}, g_old::Taylor1{TaylorN{T}}, order::Int) = g[order+1][1][1]*g_old[order+1][1][1] < zero(T)
+
+function deriv{T<:Number}(n::Int, a::Taylor1{T})
```
blas-ko: Why do you prefer this method over `derivative(n,a)`?

PerezHz: After discussing this offline with @lbenet, now I think this should co-exist with `derivative(n,a)` in TaylorSeries. Let me explain myself:

The main difference is that, as it stands now, `derivative` returns the *numerical value* of the `n`-th derivative of `a`, at the original point of expansion. But my proposal is that we should have a method (such as what I have called `deriv`) which should return the `n`-th derivative of `a` as a Taylor *polynomial*. And, only if evaluated at `zero`, it should return the *numerical value* of the `n`-th derivative of `a`, at the original point of expansion. But that's material for another PR for `TaylorSeries.jl`, which I'll submit asap!

src/surfaceintersection.jl
PerezHz: Yes @blas-ko, my original idea was that these methods should be aimed towards constructing Poincaré maps, and I'm glad that they are helping you to construct them.

But now I think that there are cases when one wants to perform only "root-finding-while-integrating", or "detect events" defined by the roots of a given scalar function `g`, without constructing a Poincaré map. That's why now I think we should have some methods for plain, general "root-finding/event-detection", separated from specialized methods for Poincaré maps.

I agree, for Poincaré maps, we should only consider "positive crossings" in the sense that you mention. And also some other things must be taken into consideration, in particular, when we want to construct Poincaré maps with jet transport techniques. Besides what you mention, some other things should be considered:

- The times of crossing may, in general, be different for each point within the initial neighbourhood (bubble) of initial conditions defined by jet transport (current root-finding methods take care of this automatically).
- For conservative flows, only initial conditions within the initial bubble defined by jet transport,  with energy equal to `E0` (i.e., the energy of the reference initial condition `q0`) should be considered. This is easy to do for Hamiltonian flows with 2 degrees of freedom (as can be seen in the material of our JuliaCon talk for the Hénon-Heiles system), but as @lbenet has pointed (offline), this is non-trivial for more than 2 degrees of freedom.

Concretely, I would suggest to add a separate file, let's say `poincare.jl`, and add there any methods related specifically to Poincaré mapping, such as the ones you propose above. Moreover, I'd propose the name `poincare` for the analogous methods of `taylorinteg` for Poincaré maps. Please, feel free to submit a PR with these changes to the `jp/surface-intersection` branch from a fork!

src/surfaceintersection.jl
PerezHz: Good catch! Thanks! Yeah, I agree that we should have a convergence criterion, I'm gonna change this part a bit and push it back here

PerezHz: Just pushed some changes which might help convergence issues, but afaict these still don't guarantee convergence of Newton-Raphson, only that it won't hang forever and it won't iterate more than necessary the NR process. If we want to *guarantee* the value found by NR, then an option is to incorporate methods from [JuliaIntervals/IntervalRootFinding.jl](https://github.com/JuliaIntervals/IntervalRootFinding.jl), particularly from `src/newton.jl`

PerezHz:
There seems to be some problems with julia 0.7-DEV when defining functions and using them within the same `@testset`; fixed it by defining the functions outside `@testset`. Now tests are passing!

PerezHz:
Rebased to current master...

lbenet:
Please note that https://github.com/JuliaDiff/TaylorSeries.jl/pull/133 has been merged, so indices of Taylor1 and TaylorN objects now start at 0; see #43.

src/surfaceintersection.jl
```diff
@@ -0,0 +1,124 @@
+# Detect if the solution crossed a root of event function g
```
lbenet: Could this be used to define a Poincare-surface? Please, add docstrings.

src/surfaceintersection.jl
```diff
@@ -0,0 +1,124 @@
...
+surfacecrossing(g_old::Taylor1{T}, g::Taylor1{T}, eventorder::Int) where {T <: Real} = g_old[eventorder+1]*g[eventorder+1] < zero(T)
+surfacecrossing(g_old::Taylor1{Taylor1{T}}, g::Taylor1{Taylor1{T}}, eventorder::Int) where {T <: Real} = g_old[eventorder+1][1]*g[eventorder+1][1] < zero(T)
+surfacecrossing(g_old::Taylor1{TaylorN{T}}, g::Taylor1{TaylorN{T}}, eventorder::Int) where {T <: Real} = g_old[eventorder+1][1][1]*g[eventorder+1][1][1] < zero(T)
+
+# An rudimentary convergence criterion for the Newton-Raphson root-finding process
+nrconvergencecriterion(g_val::T, nrabstol::T, nriter::Int, maxnriters::Int) where {T<:Real} = abs(g_val) > nrabstol && nriter ≤ maxnriters
+nrconvergencecriterion(g_val::Taylor1{T}, nrabstol::T, nriter::Int, maxnriters::Int) where {T<:Real} = abs(g_val[1]) > nrabstol && nriter ≤ maxnriters
+nrconvergencecriterion(g_val::TaylorN{T}, nrabstol::T, nriter::Int, maxnriters::Int) where {T<:Real} = abs(g_val[1][1]) > nrabstol && nriter ≤ maxnriters
+
+# This function should be moved to TaylorSeries.jl, but the definitive implementation still should be discussed
```
lbenet: If this function should be in TaylorSeries, can you submit a PR to have it there as soon as possible? I don't mind having it here, but if we can have it where it belongs, I think it is better.

src/surfaceintersection.jl
```diff
@@ -0,0 +1,124 @@
...
+    if n==0
+        return a
+    elseif n==1
+        return derivative(a)
+    else
+        return deriv(n-1, derivative(a))
+    end
+end
+
+function taylorinteg(f!, g, q0::Array{U,1}, t0::T, tmax::T,
```
lbenet: Add docstrings here to explain what is `g` and how to use it.

src/surfaceintersection.jl
```diff
@@ -0,0 +1,124 @@
...
+        order::Int, abstol::T; maxsteps::Int=500, eventorder::Int=0,
+        maxnriters::Int=5, nrabstol::T=eps(T)) where {T <: Real,U <: Number}
```
lbenet: `maxnriters=5` as default seems to me too small to be safe. I would propose at least 10 or 15. What about renaming it as `newtoniter` or something more transparent.

src/surfaceintersection.jl
```diff
@@ -0,0 +1,124 @@
...
+    const nsteps = 1
+    const nevents = 1 #number of detected events
+    const nextevord = eventorder+1
+    while t0 < tmax
+        δt_old = δt
+        δt = taylorstep!(f!, t, x, dx, xaux, t0, tmax, x0, order, abstol)
+        g_val = g(t, x, dx)
+        if surfacecrossing(g_val_old, g_val, eventorder)
+
+            #first guess: linear interpolation
```
lbenet: I guess you are looking for the value of the independent variable (t) such that `g` is zero. Why not use directly Newton-Raphson, starting at t=0? 

Also, I think that moving all the surface-crossing check to a separate function could make the code more readable.

PerezHz:
Thank you for your review, @lbenet! I will address the changes you mention!

src/surfaceintersection.jl
PerezHz: Yes, this can be used to define a Poincaré surface, but so far haven't found a way to do it automatically; it requires a bit of work from the user for each particular case; see also discussion in [this comment](https://github.com/PerezHz/TaylorIntegration.jl/pull/38#discussion_r139581165), above.

Sure, I will add docstrings! Thanks!

src/surfaceintersection.jl
PerezHz: Sure, I will add them, thanks!

src/surfaceintersection.jl
PerezHz: Sure, I'll submit such PR!

src/surfaceintersection.jl
PerezHz: I was looking for a tradeoff between speed and accuracy, but I agree with you, it's better to have a safer limit of iterations. I've just changed the value you mention, as well as the name of the variable, thanks!

src/surfaceintersection.jl
PerezHz: > I guess you are looking for the value of the independent variable (t) such that g is zero. Why not use directly Newton-Raphson, starting at t=0?

I was just looking for a "safe" starting point for the Newton-Raphson process, but yeah, that'd be another way to do it!

> Also, I think that moving all the surface-crossing check to a separate function could make the code more readable.

Nice suggestion, I'll do this right away!

src/surfaceintersection.jl
PerezHz: I was thinking of maybe adding the Hénon-Heiles Poincaré map example as a notebook to the examples folder, what do you think?

PerezHz:
Rebased to current master

PerezHz:
Just pushed a new commit which fixes a subtle bug which I found when working with non-autonomous systems (vector case); updated corresponding tests

PerezHz:
While I think it'd be nice to add the Poincaré map for the Hénon-Heiles system as an example (see [this comment](https://github.com/PerezHz/TaylorIntegration.jl/pull/38#discussion_r151605703)), perhaps I can add that in another PR, and merge this, *if* we get green lights.. what do you think, @lbenet?

lbenet:
I changed on line of a docstring. Aside from that, this is ready to be merged. I'll wait for travis to get the green flags. Thanks a lot, and sorry for the slow review...

lbenet:
There seem to be issues with travis, for Julia v0.7, but also in appveyor. 

@PerezHz Any thoughts?

src/rootfinding.jl
```diff
@@ -0,0 +1,205 @@
...
+`g_old` and `g_now` have different signs (i.e., if one is positive and the other
+one is negative); otherwise returns `false`.
+"""
+surfacecrossing(g_old::Taylor1{T}, g_now::Taylor1{T}, eventorder::Int) where {T <: Real} = g_old[eventorder]*g_now[eventorder] < zero(T)
+
+surfacecrossing(g_old::Taylor1{Taylor1{T}}, g_now::Taylor1{Taylor1{T}}, eventorder::Int) where {T <: Real} = g_old[eventorder][0]*g_now[eventorder][0] < zero(T)
+
+surfacecrossing(g_old::Taylor1{TaylorN{T}}, g_now::Taylor1{TaylorN{T}}, eventorder::Int) where {T <: Real} = g_old[eventorder][0][1]*g_now[eventorder][0][1] < zero(T)
+
+doc"""
```
ChrisRackauckas: Why Newton-Rhapson? You have a bounded 1-dimensional interval for the time where the zero is. For large equations its must faster to use an interval-based rootfinding algorithm like bisection or the Illinois algorithm. I would recommend looking at Roots.jl here.

PerezHz:
> I changed on line of a docstring. Aside from that, this is ready to be merged. 

Many thanks!

> There seem to be issues with travis, for Julia v0.7, but also in appveyor. @PerezHz Any thoughts?

I'm trying to figure out what's the issue here; for the time being, I can only say that I was able to reproduce this locally once I updated my julia 0.7 installation to 0.7.0-DEV.3244... Before that, there were no major issues with 0.7 as far as I can tell

PerezHz:
Rebased to current master... Should we add  `DiffEqBase` to the `REQUIRE` file?

src/rootfinding.jl
PerezHz: Thanks for your comment, @ChrisRackauckas! The main reason we chose Newton's method is to obtain high-accuracy answers, and Newton-Raphson gives us that at reasonable speeds. Also, the method lends itself to other numeric types, which comes pretty handy when applying e.g. jet transport techniques. I'll surely take a look at Roots.jl, thanks!

PerezHz:
@lbenet I'm trying out allowing failures in Julia nightlies... Also, what do you think about adding `DiffEqBase` to the `REQUIRE` file?

lbenet:
@PerezHz Shall I go ahead with merging this?

PerezHz:
Sure, go ahead! Many thanks for all your comments!!

lbenet:
Merged! Thanks a lot!

Maybe we should tag the current version of master

PerezHz:
I agree! I'll tag a new version right away!

PR #40, opened 2017-10-06
Update stepsize function using TaylorSeries' norm

blas-ko:
Now that `norm` is explicitely defined for `AbstractSeries` objects, `stepsize`, as it's defined, can be calculated by taking `norm(x,Inf)` for `x<:Number`, where `Number` includes all the cases for `AbstractSeries`. This PR adressess this. `src/jettransport.jl` was removed as it was no longer needed; nevertheless, `test/jettransport.jl` tests still do their job correctly.


lbenet:
I like the idea; however, tests are not passing... I think the problem is that now, some methods of `stepsize` are not found/handled correctly.

blas-ko:
Tests pass locally on my computer. I'm using `TaylorSeries.jl` `master` branch. I don't know if `TaylorIntegration.jl` uses `TaylorSeries`' last tag release or last commit (`master`); but seeing travis' error log, I think it's the first case, as the `issless` problem is due to `norm(a,p) <: AbstractSeries` and not a non-negative number.

PerezHz:
This is cool stuff! Thanks @blas-ko! I can confirm, that using latest TaylorSeries master locally, tests for this PR pass...

Now, for ci tests, travis and appveyor use the latest tagged version of required packages. For the time being, a possible fix is to add a `Pkg.checkout("TaylorSeries")` in both `.travis.yml` and `appveyor.yml` at the right places: try substituting, respectively, line 14 of `.travis.yml` with the following line:

```
 - julia -e 'Pkg.clone(pwd()); Pkg.checkout("TaylorSeries"); Pkg.build("TaylorIntegration"); Pkg.test("TaylorIntegration"; coverage=true)'
```
(note the whitespace at the beginning of line)

and line 31 of `appveyor.yml` with:

```
      Pkg.checkout(\"TaylorSeries\"); Pkg.clone(pwd(), \"TaylorIntegration\"); Pkg.build(\"TaylorIntegration\")"
```
(note the whitespaces here, too)

And then, once a new TaylorSeries.jl version is tagged, we can get rid of this temporary hacks! We've done this before, e.g., for #18

lbenet:
As @PerezHz mentions, the last tagged version, in this case, TaylorSeries 0.5.0, is used. My proposal is get https://github.com/JuliaDiff/TaylorSeries.jl/pull/121, tag a new version (the last one compatible with Julia 0.5, and then come back to this PR.

PerezHz:
Another comment: part of the reason of having separate `stepsize` methods for jet transport, is that we were trying to implement Pérez-Palau et al. formula (see page 8 of [this paper](https://upcommons.upc.edu/bitstream/handle/2117/85050/preprint-LCS.pdf)) in order to ensure that *all* orders of the expansion (both wrt time and the variations) were within the absolute tolerance, `abstol`... For low-order jet transport, such as the examples included in the tests, the difference may be negligible; but for higher orders, this may become relevant. So we'd have to take a careful look to see if the new methods of calculating the time step for jet transport integrations indeed follow Pérez-Palau formula.

lbenet:
Testing specific systems is the way to go.

blas-ko:
>  we'd have to take a careful look to see if the new methods of calculating the time step for jet transport integrations indeed follow Pérez-Palau formula.

jt-sz

Note, from the paper, that `ψ_k^(i) = max_m |c_m,k^(i)|`. When he takes `h_opt`, he takes `min_k { (a/ψ_k^b)^(1/(b-1) }`, which corresponds to `max_i { max_k { ψ_k^(i) } } == max_i { norm(x^(i), Inf) } }` for `i ∈ [N,N-1]`. 
Indeed, "our" method is identical to Pérez-Palau's formula (except, of course, that we're using absolute error and he's using relative, but that was stated this way before the PR).

lbenet:
Maybe we should include the possibility to use a relative error (instead of the absolute one) to determine the step size; but this should be another issue.

blas-ko:
Let's try to restart travis now that `TaylorSeries` v0.51 is out!

PerezHz:
Just restarted them; travis on julia 0.6 seems to be passing!

src/TaylorIntegration.jl
```diff
@@ -8,7 +8,7 @@ export taylorinteg, liap_taylorinteg
 
 include("explicitode.jl")
 
-include("jettransport.jl")
+#include("jettransport.jl")
```
PerezHz: Go ahead and remove this line

src/explicitode.jl
```diff
@@ -103,12 +103,13 @@ function stepsize{T<:Real, U<:Number}(x::Taylor1{U}, epsilon::T)
...
     h = convert(T, Inf)
     for k in (ord-1, ord)
-        @inbounds aux = abs( x[k+1] )
+        @inbounds aux = norm( x[k+1], Inf)
         aux == zero(T) && continue
         aux = epsilon / aux
         kinv = one(T)/k
         aux = aux^kinv
         h = min(h, aux)
+        h
```
PerezHz: This line should also be removed?

src/TaylorIntegration.jl
blas-ko: done!

src/explicitode.jl
blas-ko: done too! Actually I don't know how that line got there xD.

src/TaylorIntegration.jl
PerezHz: 👌

src/explicitode.jl
PerezHz: Great!

PerezHz:
Awesome work @blas-ko! Thanks for your contribution! I think this is ready for merging, what do you think @lbenet?

lbenet:
LGTM!! Very nice that the addition of 3 lines saves so many! I am in favor of merging this.

Issue #19, opened 2017-02-23
Add JuliaDiffEq Common Interface Bindings


ChrisRackauckas:
Hey,
  This package has really evolved and looks really cool! I was thinking it would be nice to add common interface bindings. To do this, it would just need to add a dependency on DiffEqBase and implement a `solve` function which translates how to solve an `ODEProblem` to the solver for this package. 

Taking a quick look at the examples, it looks like it just would extract the information of the `ODEProblem` (the function, timespan, and initial condition), and use it to build the

```julia
taylorinteg(kepler_problem, q0, t0, 0.01, order, abs_tol, maxsteps)
```

command. The algorithm could be

```julia
alg = TaylorIntegration(order=SomeDefaultHere)
```

And then

```julia
solve(prob,alg,maxiters=...,abstol=...)
```

would be a solve command that translates the arguments and calls `taylorinteg` appropriately. Would you accept such a PR?

Now, the big issue that I see from the examples is that they're not using a inplace format. Your examples seem to use the ODE.jl `du=f(t,u)` version, but doesn't include support for `f(t,u,du)`. Is that correct? If so, that hampers performance quite a bit, but we can translate any inplace function to the not-inplace form using the same translation as the ODE.jl common interface bindings.

ChrisRackauckas:
How exactly are the timesteps chosen?

lbenet:
Thanks for the comments and the interest! We surely accept PRs!

We have discussed (offline) about adapting our package for the inplace evaluation, just as you have it in `DifferentialEquations.jl`. @PerezHz did some benchmarks and I think we were getting ~%20 improvement. Maybe that, combined with #18, could yield even better results. This is something we are on right now. 

Yet, `TaylorIntegration` is comparing poorly in terms of speed with other methods due to memory allocations, which at the end is due to the way `TaylorSeries.jl` is constructed. One way to go around that is to construct a macro which parses `f(t, u)` (or `f(t, u, du)`) in such a way that we get a function that implements the equations in terms of the Taylor coefficients,; we haven't got anything interesting yet... Also, after some discussions with @dpsanders, we are reconsidering the design of `TaylorSeries.jl`, to make at least such parsing easier.

[...]

PR #44, opened 2017-12-18
add common interface bindings

ChrisRackauckas:
This adds common interface bindings (#19) so that 

```julia
sol = solve(prob,TaylorMethod(50))
```

calls the Taylor method from this package with maximum order set to 50. There's definitely more that can be done integrating the two packages, but this is a good and usable first step.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -62,8 +62,6 @@ function DiffEqBase.solve{uType,tType,isinplace,AlgType<:TaylorAlgorithm}(
       _timeseries = vec(vectimeseries)
     end
 
-    _t,_timeseries
-
     build_solution(prob,  alg, _t, _timeseries,
                    timeseries_errors = timeseries_errors,
                    retcode = :Success)
--- a/src/common.jl
+++ b/src/common.jl
@@ -1,0 +1,70 @@
+abstract type TaylorAlgorithm <: DEAlgorithm end
+struct TaylorMethod <: TaylorAlgorithm
+    order::Int
+end
+
+TaylorMethod() = error("Maximum order must be specified for the Taylor method")
+
+export TaylorMethod
+
+function DiffEqBase.solve{uType,tType,isinplace,AlgType<:TaylorAlgorithm}(
+    prob::AbstractODEProblem{uType,tType,isinplace},
+    alg::AlgType,
+    timeseries=[],ts=[],ks=[];
+    verbose=true, abstol = 1e-6, save_start = true,
+    timeseries_errors=true, maxiters = 1000000,
+    callback=nothing, kwargs...)
+
+    if verbose
+        warned = !isempty(kwargs) && check_keywords(alg, kwargs, warnlist)
+        warned && warn_compat()
+    end
+
+    if prob.callback != nothing || callback != nothing
+        error("TaylorIntegration is not compatible with callbacks.")
+    end
+
+    if typeof(prob.u0) <: Number
+        u0 = [prob.u0]
+    else
+        u0 = vec(deepcopy(prob.u0))
+    end
+
+    sizeu = size(prob.u0)
+
+    if !isinplace && (typeof(prob.u0)<:Vector{Float64} || typeof(prob.u0)<:Number)
+        f! = (t, u, du) -> (du .= prob.f(t, u); 0)
+    elseif !isinplace && typeof(prob.u0)<:AbstractArray
+        f! = (t, u, du) -> (du .= vec(prob.f(t, reshape(u, sizeu))); 0)
+    elseif typeof(prob.u0)<:Vector{Float64}
+        f! = prob.f
+    else # Then it's an in-place function on an abstract array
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
Need to remember to merge doc change https://github.com/JuliaDiffEq/DiffEqDocs.jl/pull/64 when this merges.

ChrisRackauckas:
Tests fail on nightlies. I usually set nightlies as allowed failures to avoid handling the volatility.

lbenet:
Thanks for contributing @ChrisRackauckas. I have a busy morning; I'll try to take a look on this in the afternoon.

src/common.jl
```diff
@@ -0,0 +1,70 @@
...
+
+TaylorMethod() = error("Maximum order must be specified for the Taylor method")
+
+export TaylorMethod
+
+function DiffEqBase.solve{uType,tType,isinplace,AlgType<:TaylorAlgorithm}(
+    prob::AbstractODEProblem{uType,tType,isinplace},
+    alg::AlgType,
+    timeseries=[],ts=[],ks=[];
+    verbose=true, abstol = 1e-6, save_start = true,
```
lbenet: I suggest to have `abstol = 1e-20` as default value, or smaller, to actually get the accuracy that the method allows. 

`abstol` is used to define the step-size in each iteration, and corresponds to the size of the last (or one-before-last) term of the local Taylor expansion. Having it very small is what makes the error (per integration step) to be the round-off error.

src/common.jl
```diff
@@ -0,0 +1,70 @@
...
+    if typeof(prob.u0) <: AbstractArray
+      _timeseries = Vector{uType}(0)
+      for i=start_idx:size(vectimeseries, 1)
+          push!(_timeseries, reshape(view(vectimeseries, i, :, )', sizeu))
+      end
+    else
+      _timeseries = vec(vectimeseries)
+    end
+
+    _t,_timeseries
```
lbenet: Sorry for the question, but what is the aim of this line?

test/common.jl
```diff
@@ -0,0 +1,21 @@
+using TaylorIntegration, Base.Test
+
+@testset "Test integration of ODE with numbers in common interface" begin
+    f(t,u) = u
+    u0 = 0.5
+    tspan = (0.0,1.0)
+    prob = ODEProblem(f,u0,tspan)
+    sol = solve(prob,TaylorMethod(50))
+
+    @test sol[end] - 0.5exp(1) < 1e-12
```
lbenet: If the default value for `abstol` is set to `1e-20`, this test can check the equality!

lbenet:
> Tests fail on nightlies. I usually set nightlies as allowed failures to avoid handling the volatility.

I have also noticed this, though I do not understand the problem. In #31 I changed travis in the way you propose, but it is perhaps better to have that merged now, together with this PR.

src/common.jl
ChrisRackauckas: oh it was from testing.

src/common.jl
ChrisRackauckas: In here:

http://docs.juliadiffeq.org/latest/basics/common_solver_opts.html#Basic-Stepsize-Control-1

it states that the universal default (for ODEs at least) is `abstol=1e-6`. This is what every other package (even the old C++ and Fortran ones) tend to do anyways, so it has precedence. In the docs part mentioning this package I mentioned it should be used for high accuracy calculations, but I think that it should keep the common default just for consistency.

lbenet: I get your point, and agree that a comment in the docs is always worth, and what you wrote is quite clear on that.

Yet, I prefer to have "useful" defaults (as user) rather than standardized ones. Having `abstol = 1e-6` makes me specifying it every time (because I use `TaylorIntegration` for accurate results); having it as `1e-20` allows me to skip it, unless I want even more accuracy.

test/common.jl
lbenet: I just checked that equality also holds with `abstol = 1e-6`.

src/common.jl
ChrisRackauckas: It is a useful default though. Most people solving an ODE don't want or need `abstol = 1e-6`. Only in the case of accurate solving does something like `absol = 1e-20` make sense. `absol = 1e-20` could never be the global default since it would also just break in a lot of cases due to floating point error.

lbenet: Just a quick comment: I agree with you that for most people `abstol = 1e-6` is enough, but they will not be using `TaylorIntegration.jl`. If they use `TaylorIntegration.jl` is because they do want to push the accuracy to its limits, and in that case having `abstol = 1e-6` will generate a step-size that is probably so large that the results are not so close to the exact solution. That's the reason I think that `abstol = 1e-20` is reasonable for TaylorIntegration, whereas for other used methods a reasonable default is `1e-6`, as it is already.

lbenet:
Thanks a lot again for this nice addition!

PerezHz:
This is really cool!! Thanks for adding this, @ChrisRackauckas!

PR #47, opened 2018-02-19
Update TaylorSeries required version to 0.7.1

PerezHz:
@lbenet what do you think about updating the required version of TaylorSeries to v0.7.1 on master?


lbenet:
Thanks for noting this! Merging.

PR #48, opened 2018-02-28
Some improvements in Lyapunov spectrum computation and other minor stuff

PerezHz:
This PR adds non-breaking changes to internal Lyapunov spectrum code, and should improve performance a bit. In particular, `stabilitymatrix!` might not be necessary anymore. Tests are passing in julia 0.6.

Also, this PR `@reexport`s TaylorSeries (this change is also included also in #31, but it is really helpful! so I thought it'd be nice to have this already in current master, while #31 is merged)

cc: @lbenet

### DIFF
--- a/src/liapunovspectrum.jl
+++ b/src/liapunovspectrum.jl
@@ -160,8 +160,11 @@ end
     liap_taylorinteg(f, q0, t0, tmax, order, abstol; maxsteps::Int=500)
 
 Similar to [`taylorinteg!`](@ref) for the calculation of the Liapunov
-spectrum. Whenever `length(q0) != TaylorSeries.get_numvars()`, then
-`liap_taylorinteg` changes globally some internal parameters of TaylorN.
+spectrum. Note that the number of `TaylorN` variables should be set
+previously by the user (e.g., by means of `TaylorSeries.set_variables`) and
+should be equal to the length of the vector of initial conditions `q0`.
+Otherwise, whenever `length(q0) != TaylorSeries.get_numvars()`, then
+`liap_taylorinteg` throws an `AssertionError`.
 
 """
 function liap_taylorinteg(f, q0::Array{U,1}, t0::T, tmax::T,
@@ -175,13 +178,10 @@ function liap_taylorinteg(f, q0::Array{U,1}, t0::T, tmax::T,
     const jt = eye(U, dof)
     const _δv = Array{TaylorN{Taylor1{U}}}(dof)
 
-    if get_numvars() == dof
-        for ind in eachindex(q0)
-            _δv[ind] = TaylorN(Taylor1{U},ind,order=1)
-        end
-    else
-        # NOTE: This changes GLOBALLY internal parameters of TaylorN
-        _δv = set_variables(Taylor1{U}, "δ", order=1, numvars=dof)
+    @assert get_numvars() == dof "`length(q0)` must be equal to number of variables set by `TaylorN`"
+
+    for ind in eachindex(q0)
+        _δv[ind] = TaylorN(Taylor1{U},ind,order=1)
     end
 
     # Initial conditions
@@ -263,13 +263,10 @@ function liap_taylorinteg(f, q0::Array{U,1}, trange::Union{Range{T},Vector{T}},
     const jt = eye(U, dof)
     const _δv = Array{TaylorN{Taylor1{U}}}(dof)
 
-    if get_numvars() == dof
-        for ind in eachindex(q0)
-            _δv[ind] = TaylorN(Taylor1{U},ind,order=1)
-        end
-    else
-        # NOTE: This changes GLOBALLY internal parameters of TaylorN
-        _δv = set_variables(Taylor1{U}, "δ", order=1, numvars=dof)
+    @assert get_numvars() == dof "`length(q0)` must be equal to number of variables set by `TaylorN`"
+
+    for ind in eachindex(q0)
+        _δv[ind] = TaylorN(Taylor1{U},ind,order=1)
     end
 
     # Initial conditions
--- a/src/liapunovspectrum.jl
+++ b/src/liapunovspectrum.jl
@@ -9,13 +9,13 @@ at `x`; `x` is of type `Vector{T<:Number}`. `δx` and `dδx` are two
 auxiliary arrays of type `Vector{TaylorN{T}}` to avoid allocations.
 
 """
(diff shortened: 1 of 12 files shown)


PerezHz:
The philosophy of this change was originally to make `jetcoeffs!` and `liap_jetcoeffs!` as similar as possible, with a `@liap_taylorize_ode` macro in mind 😄, but then I saw some inconsistencies in the code, and one thing led to another, and it was clear that we could improve performance a bit!

lbenet:
I'll try to take a look on this tonight.... sorry, lots of nonsensical-admin tasks.

src/liapunovspectrum.jl
```diff
@@ -171,7 +173,7 @@ function liap_taylorinteg(f, q0::Array{U,1}, t0::T, tmax::T,
     const jt = eye(U, dof)
 
     # NOTE: This changes GLOBALLY internal parameters of TaylorN
-    global _δv = set_variables("δ", order=1, numvars=dof)
+    global _δv = set_variables(Taylor1{U}, "δ", order=1, numvars=dof)
```
lbenet: We should add a clause in the docstrings about this global change.

src/liapunovspectrum.jl
```diff
@@ -98,26 +98,28 @@ function liap_jetcoeffs!(eqsdiff!, t::Taylor1{T}, x::Vector{Taylor1{U}},
...
         # Set `xaux`, auxiliary vector of Taylor1 to order `ord`
         for j in eachindex(x)
             @inbounds xaux[j] = Taylor1( x[j].coeffs[1:ordnext] )
         end
 
         # Equations of motion
-        eqsdiff!(taux, xaux, dx)
-        stabilitymatrix!( eqsdiff!, t[0], view(xaux,1:dof), δx, dδx, jac )
+        for ind in eachindex(δx)
+            @inbounds δx[ind] = xaux[ind] + _δv[ind]
```
lbenet: What is `_δv`? I guess it is some global variable; if so, maybe it is better to pass it as an argument in the function call.

This for loop may be merged with the one above (`x` and `δx` have the same length).

src/liapunovspectrum.jl
```diff
@@ -98,26 +98,28 @@ function liap_jetcoeffs!(eqsdiff!, t::Taylor1{T}, x::Vector{Taylor1{U}},
...
+        end
+        eqsdiff!(taux, δx, dδx)
+        @inbounds dx[1:dof] .= constant_term.(dδx)
+        # Stability matrix
```
lbenet: I do see why this approach leads to some improvement; you avoid certain allocations in `stabilitymatrix!` related to the independent variables that are represented by `_δv` now. I think it would be nice to still have `stabilitymatrix!` being called, maybe by passing to it `_δv` as a new argument, but avoiding allocations the way you do it.

src/liapunovspectrum.jl
lbenet: Thinking it over, maybe we should rewrite this checking if the current internal parameters of TaylorN allow to skip this redefinition, so we define `_δv` is simply the "order one" independent variables. (Maybe the user cares about having a larger order.) To do this, we should check that the number of variables is compatible. This is only a proposal, which may be thrown away if it does not improve performance and allocations.

PerezHz: Thanks for the suggestion! I'll try it out!

src/liapunovspectrum.jl
PerezHz: Exactly, `_δv` is the global variable declared in `liap_taylorinteg`; I also thought of merging the loops but actually `δx` has the length of the vector of initial conditions (say, `q0`) whereas `x` has the length of `q0` plus the size of the stability matrix

PerezHz: I'll try to pass `_δv` as an argument and see if performance improves

src/liapunovspectrum.jl
PerezHz: Also, we were evaluating twice the equations of motion per timestep per order; with this change now we are evaluating only once the equations of motion per timestep per order.

src/liapunovspectrum.jl
PerezHz: I also was thinking of adding a warning whenever the `_δv` are set inside `liap_taylorinteg`? What do you think?

src/liapunovspectrum.jl
lbenet: (I actually noticed this but still added the comment)

Ok, with the two separated loops.

src/liapunovspectrum.jl
lbenet: Nice!

lbenet:
Let me know when this is ready, or simply go ahead and merge it!

PerezHz:
Thank you for all your comments, @lbenet! Just pushed some changes which address the suggestions you made. Also, I updated the required version of TaylorSeries to v0.7.2

lbenet:
Did you checked the performance?

PerezHz:
Yes I did; performance did improve a bit after the last change!

src/liapunovspectrum.jl
```diff
@@ -169,9 +173,16 @@ function liap_taylorinteg(f, q0::Array{U,1}, t0::T, tmax::T,
...
     const jt = eye(U, dof)
+    const _δv = Array{TaylorN{Taylor1{U}}}(dof)
 
-    # NOTE: This changes GLOBALLY internal parameters of TaylorN
-    global _δv = set_variables("δ", order=1, numvars=dof)
+    if get_numvars() == dof
+        for ind in eachindex(q0)
+            _δv[ind] = TaylorN(Taylor1{U},ind,order=1)
+        end
+    else
```
lbenet: A question here: if the `else` applies, should we throw an error (DomainError, or something like that)? My point
is that the user should be responsible of this subtlety...

Same comment applies below.

lbenet:
There seem to be some problems...

PerezHz:
Just checked the coverage; gotta add a test precisely for the `else` case you mentioned, and I agree that we should add some kind of error message, or at the very least a warning

PerezHz:
Just pushed a commit which solves the coverage issue.. Regarding the error which we discussed should be thrown, I opted for the `@assert` / `AssertionError`, so whenever things don't match `liap_taylorinteg` simply throws this error.

lbenet:
LGTM. Merging! Thanks a lot!

PerezHz:
Thanks for merging! 😄

PR #49, opened 2018-03-31
Update the common interface bindings

ChrisRackauckas:


PerezHz:
Thanks, Chris! Is this ready to be merged or do you want to add some more changes?

@lbenet : do you think it would be worth to tag a new TaylorIntegration version (v0.2.2; current is v0.2.1) after merging this, since changes included in DiffEqBase v3.8.0 are breaking?

ChrisRackauckas:
It would require a minor (0.3.0) because of the minimum version bound change.

>Is this ready to be merged or do you want to add some more changes?

I don't think there's anything else to do here.

PerezHz:
> It would require a minor (0.3.0) because of the minimum version bound change.

You're right; I agree

> I don't think there's anything else to do here.

Ok, great! I'll merge right away!

PerezHz:
Again, many thanks, Chris!

PR #50, opened 2018-04-05
Reduce closure size

ChrisRackauckas:
Maybe this fixes the issue in #31 by making sure `prob` won't be in the closure?

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -32,15 +32,16 @@ function DiffEqBase.solve{uType,tType,isinplace,AlgType<:TaylorAlgorithm}(
 
     sizeu = size(prob.u0)
     p = prob.p
+    f = prob.f
 
     if !isinplace && (typeof(prob.u0)<:Vector{Float64} || typeof(prob.u0)<:Number)
-        f! = (t, u, du) -> (du .= prob.f(u, p, t); 0)
+        f! = (t, u, du) -> (du .= f(u, p, t); 0)
     elseif !isinplace && typeof(prob.u0)<:AbstractArray
-        f! = (t, u, du) -> (du .= vec(prob.f(reshape(u, sizeu), p, t)); 0)
+        f! = (t, u, du) -> (du .= vec(f(reshape(u, sizeu), p, t)); 0)
     elseif typeof(prob.u0)<:Vector{Float64}
-        f! = (t, u, du) -> prob.f(du, u, p, t)
+        f! = (t, u, du) -> f(du, u, p, t)
     else # Then it's an in-place function on an abstract array
-        f! = (t, u, du) -> (prob.f(reshape(du, sizeu), reshape(u, sizeu), p, t);
+        f! = (t, u, du) -> (f(reshape(du, sizeu), reshape(u, sizeu), p, t);
                             u = vec(u); du=vec(du); 0)
     end
 


ChrisRackauckas:
@lbenet

lbenet:
Thanks @ChrisRackauckas 

I just restarted travis (Julia 0.6) because something odd happened. In any case, tests in Julia 0.6 seem to pass. 

One question: which cases (of the `if` block in lines 36-46) are actually covered by the two tests in test/common.jl? Yesterday, playing with this I built some sort of function barrier, which was helpful, but also noticed that sometimes a case was recognized by the first conditional, and sometimes by the last. (Other odd things also happened, so i dodn't pushed my solution....)

lbenet:
Yesterday, while I was playing on this, I moved out the `if` block to a (new) function which was returning the anonymous call, and using the same trick as in here. That, and twisting a bit the last case, made the tests for the common interface to pass. But then a test in `jetttransport.jl` was broken; I truly do not understand this...

ChrisRackauckas:
What about the function are you actually introspecting/using?

lbenet:
I don't I understand your question. 

What I did simply was to move the `if` block in `DiffEqsBase.solve`, to a function `_to_taylorinteg_f(prob::AbstractODEProblem)` which returned the anonymous function that you had in the block. (Some other minor things I changed were to return `nothing` instead of `0`, define `ff = prob.f` and use some auxiliary variables for the `else` part.)

That solved the problems that I mentioned in #31 related to `test/common.jl`, but *strangely*, it throws an error in `test/jettransport.jl`, which is quite unexpected. This is the beginning of what i get:
```julia
Test Taylor1 jet transport (t0, tmax): 1-dim case |   54     54
ERROR: LoadError: LoadError: MethodError: no method matching convert(::Type{AssertionError}, ::String)
Closest candidates are:
  convert(::Type{Any}, ::ANY) at essentials.jl:28
  convert(::Type{T}, ::T) where T at essentials.jl:29
Stacktrace:
 [1] AssertionError(::String) at ./coreimg.jl:14
 [2] update_valid_age!(::UInt64, ::UInt64, ::Core.Inference.InferenceState) at ./inference.jl:2353
[...]
```

Moving the location of the `common.jl` tests (as the first file in `runtests.jl` remedies everything). 

This is quite unexpected, and I dislike this behavior because there should be no dependence of src/common.jl` or `src/parse_eqs.jl` in what already was working.

lbenet:
This is the code of the function containing the `if` block; inside solve, it was simply called as `f! = _to_taylorinteg_f(prob)`.

```julia
function _to_taylorinteg_f(prob::AbstractODEProblem)

    sizeu = size(prob.u0)
    iiplace = isinplace(prob)
    p = prob.p
    ff = prob.f

    if !iiplace && (typeof(prob.u0)<:Vector{Float64} || typeof(prob.u0)<:Number)
        return (t, u, du) -> (du .= ff(u, p, t); nothing)
    elseif !iiplace && typeof(prob.u0)<:AbstractArray
        return (t, u, du) -> (du .= vec(ff(reshape(u, sizeu), p, t));
                            nothing)
    elseif typeof(prob.u0)<:Vector{Float64}
        return (t, u, du) -> (ff(du, u, p, t); nothing)
    else # Then it's an in-place function on an abstract array
        return (t, u, du) -> (v = reshape(u, sizeu); dv = reshape(du, sizeu);
                            ff(dv, v, p, t);
                            u = vec(deepcopy(v)); du=vec(deepcopy(dv));
                            nothing)
    end
end
```

ChrisRackauckas:
That's just because of this:

https://github.com/PerezHz/TaylorIntegration.jl/blob/master/test/jettransport.jl#L9-L10

This is not in a testset block so it's globally defined and not cleared after a test.

lbenet:
Good catch! I'll change it and rerun the tests.

lbenet:
I moved inside the `@testset` blocks the definitions of `f` and `g`; the problem persists when I test `runtests.jl`, with the same error message. Individually, both `jettransport.jl` and `common.jl` succeed.

ChrisRackauckas:
Make common use a function name other than `f` and see what happens?

lbenet:
In the tests, or in common.jl?

ChrisRackauckas:
In the tests. Inside it's using anonymous functions which won't have namespacing issues.

lbenet:
Same problem...

lbenet:
Again, the strange thing is that if I include *first* the `common.jl` tests, and then the rest (also with `common.jl` where it is), tests pass...

lbenet:
The problem in #31 seems related to this, though it remains; see the discussion in #31. I merge this, since it contributes to readability.

Thanks @ChrisRackauckas

PR #54, opened 2018-09-05
Include version upper bounds for requirements

lbenet:
This PR is to prepare the transition to support Julia 0.7 and Julia 1.0; I suggest to merge this (and other commits), release patch a version, which would be the last to support Julia 0.6.


PerezHz:
Great, I agree! I'll just go ahead and merge this, thanks!

PR #55, opened 2018-09-08
Lyapunov spectrum: Add support for user-defined Jacobians

PerezHz:
This PR adds support for user-defined Jacobians in Lyapunov spectrum computations, as an additional alternative to computing it via automatic differentiation with TaylorSeries, as is currently done. This improves substantially execution times. As an example, for the Lorenz system, current usage is the following:

```julia
using TaylorIntegration

order = 28
abstol = 1.0E-20
q0 = [19.0, 20.0, 50.0] #the initial condition
t0 = 0.0 #the initial time
tmax = t0+20.0 #final time of integration
xi = set_variables("δ", order=1, numvars=length(q0)) #Initialize TaylorN variables for autodiff computation of Jacobian

#Lorenz system parameters
const σ = 16.0
const β = 4.0
const ρ = 45.92

#Lorenz system ODE:
function lorenz!(t, x, dx)
    dx[1] = σ*(x[2]-x[1])
    dx[2] = x[1]*(ρ-x[3])-x[2]
    dx[3] = x[1]*x[2]-β*x[3]
    nothing
end

@time tv, xv, λv = lyap_taylorinteg(lorenz!, q0, t0, tmax, order, abstol; maxsteps=2000)
```

for which a typical output on my machine is (2nd run):
`1.728911 seconds (41.31 M allocations: 2.764 GiB, 13.65% gc time)`

Using this PR, if we define
```julia
#Lorenz system Jacobian (in-place):
function lorenz_jac!(jac, t, x, dx)
    jac[1,1] = -σ+0.0*x[1]; jac[1,2] = σ+0.0*x[1]; jac[1,3] = 0.0*x[1]
    jac[2,1] = ρ-x[3]; jac[2,2] = -1.0+0.0*x[1]; jac[2,3] = -x[1]
    jac[3,1] = x[2]; jac[3,2] = x[1]; jac[3,3] = -β+0.0*x[1]
    nothing
end
```
i.e., a function which evaluates the Lorenz system Jacobian in-place, then we can do:

```julia
@time tv, xv, λv = lyap_taylorinteg(lorenz!, q0, t0, tmax, order, abstol, lorenz_jac!; maxsteps=2000)
```

which on my machine outputs (2nd run):

`0.347845 seconds (4.24 M allocations: 447.103 MiB, 6.86% gc time)`

Note that the Jacobian function `lorenz_jac` is added as an optional argument to the `lyap_taylorinteg` call. Providing an user-defined Jacobian avoids the use of `TaylorN` altogether (e.g., in the example below, the line where the `set_variables` function is called, is not necessary if the user provides the Jacobian. Main performance gains come from avoiding intensive use of `TaylorN`, which are computationally expensive to manipulate.

Corresponding tests for the new functionality have been added to this PR.

cc: @lbenet

### DIFF
--- a/src/lyapunovspectrum.jl
+++ b/src/lyapunovspectrum.jl
@@ -189,7 +189,7 @@ function lyap_taylorinteg(f, q0::Array{U,1}, t0::T, tmax::T,
 
     # Initial conditions
     @inbounds tv[1] = t0
-    for ind in 1:dof
+    for ind in eachindex(q0)
         @inbounds xv[ind,1] = q0[ind]
         @inbounds λ[ind,1] = zero(U)
         @inbounds λtsum[ind] = zero(U)
@@ -232,7 +232,7 @@ function lyap_taylorinteg(f, q0::Array{U,1}, t0::T, tmax::T,
         tspan = t0-t00
         nsteps += 1
         @inbounds tv[nsteps] = t0
-        @inbounds for ind in 1:dof
+        @inbounds for ind in eachindex(q0)
             xv[ind,nsteps] = x0[ind]
             λtsum[ind] += log(RH[ind,ind])
             λ[ind,nsteps] = λtsum[ind]/tspan
@@ -274,7 +274,7 @@ function lyap_taylorinteg(f, q0::Array{U,1}, trange::Union{AbstractRange{T},Vect
     end
 
     # Initial conditions
-    @inbounds for ind in 1:dof
+    @inbounds for ind in eachindex(q0)
         xv[ind,1] = q0[ind]
         λ[ind,1] = zero(U)
         λtsum[ind] = zero(U)
@@ -319,7 +319,7 @@ function lyap_taylorinteg(f, q0::Array{U,1}, trange::Union{AbstractRange{T},Vect
             t0 += δt
             @inbounds t[0] = t0
             nsteps += 1
-            @inbounds for ind in 1:dof
+            @inbounds for ind in eachindex(q0)
                 λtsum[ind] += log(RH[ind,ind])
             end
             for ind in eachindex(QH)
@@ -338,7 +338,7 @@ function lyap_taylorinteg(f, q0::Array{U,1}, trange::Union{AbstractRange{T},Vect
         end
         iter += 1
         tspan = t0-t00
-        @inbounds for ind in 1:dof
+        @inbounds for ind in eachindex(q0)
             xv[ind,iter] = x0[ind]
             λ[ind,iter] = λtsum[ind]/tspan
         end
--- a/src/lyapunovspectrum.jl
+++ b/src/lyapunovspectrum.jl
@@ -259,8 +259,7 @@ function lyap_taylorinteg(f, q0::Array{U,1}, trange::Union{AbstractRange{T},Vect
     fill!(xv, U(NaN))
     λ = similar(xv)
     λtsum = similar(q0)
-    # jt = eye(U, dof)
-    jt = Matrix{U}(I, dof, dof)#eye(U, dof)
+    jt = Matrix{U}(I, dof, dof)
     _δv = Array{TaylorN{Taylor1{U}}}(undef, dof)
 
     @assert get_numvars() == dof "`length(q0)` must be equal to number of variables set by `TaylorN`"
@@ -319,7 +318,7 @@ function lyap_taylorinteg(f, q0::Array{U,1}, trange::Union{AbstractRange{T},Vect
             for ind in eachindex(QH)
                 @inbounds x0[dof+ind] = QH[ind]
             end
-            x .= Taylor1.( x0, order )
+            @inbounds x .= Taylor1.( x0, order )
             t0 ≥ t1 && break
         end
         if nsteps ≥ maxsteps && t0 != t1
--- a/src/lyapunovspectrum.jl
+++ b/src/lyapunovspectrum.jl
@@ -97,33 +97,40 @@ and `xaux`, `δx` and `dδx` are auxiliary vectors.
 function lyap_jetcoeffs!(eqsdiff!, t::Taylor1{T}, x::Vector{Taylor1{U}},
         dx::Vector{Taylor1{U}}, xaux::Vector{Taylor1{U}},
         δx::Array{TaylorN{Taylor1{U}},1}, dδx::Array{TaylorN{Taylor1{U}},1},
-        jac::Array{Taylor1{U},2}, _δv::Array{TaylorN{Taylor1{U}}}) where {T<:Real, U<:Number}
+        jac::Array{Taylor1{U},2}, _δv::Array{TaylorN{Taylor1{U}}}, eqsdiff_jac! =Nothing) where {T<:Real, U<:Number}
     order = x[1].order
     # Dimensions of phase-space: dof
     nx = length(x)
     dof = length(δx)
     for ord in 0:order-1
         ordnext = ord+1
-
         # Set `taux`, auxiliary Taylor1 variable to order `ord`
         @inbounds taux = Taylor1( t.coeffs[1:ordnext] )
         # Set `xaux`, auxiliary vector of Taylor1 to order `ord`
         for j in eachindex(x)
             @inbounds xaux[j] = Taylor1( x[j].coeffs[1:ordnext] )
         end
-        # Set δx equal to current value of xaux plus 1st-order variations
-        for ind in eachindex(δx)
-            @inbounds δx[ind] = xaux[ind] + _δv[ind]
+
+        if eqsdiff_jac! == Nothing
+            # Set δx equal to current value of xaux plus 1st-order variations
+            for ind in eachindex(δx)
+                @inbounds δx[ind] = xaux[ind] + _δv[ind]
+            end
+            # Equations of motion
+            # TODO! define a macro to optimize the eqsdiff
+            eqsdiff!(taux, δx, dδx)
+            @inbounds dx[1:dof] .= constant_term.(dδx)
+            # Stability matrix
+            jacobian!(jac, dδx)
+        else
+            # Equations of motion
+            # TODO! define a macro to optimize the eqsdiff
+            eqsdiff!(taux, xaux, dx)
+            # Stability matrix
+            eqsdiff_jac!(jac, taux, xaux, dx)
         end
 
-        # Equations of motion
-        # TODO! define a macro to optimize the eqsdiff
-        eqsdiff!(taux, δx, dδx)
-        @inbounds dx[1:dof] .= constant_term.(dδx)
-        # Stability matrix
-        jacobian!(jac, dδx)
         @inbounds dx[dof+1:nx] = jac * reshape( xaux[dof+1:nx], (dof,dof) )
-
         # Recursion relations
         for j in eachindex(x)
             @inbounds x[j][ordnext] = dx[j][ord]/ordnext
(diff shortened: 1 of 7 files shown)


src/lyapunovspectrum.jl
```diff
@@ -87,42 +87,56 @@ function modifiedGS!(A, Q, R, aⱼ, qᵢ, vⱼ)
...
+(i.e., the Jacobian), and `xaux`, `δx`, `dδx` and `_δv` are auxiliary vectors.
+Optionally, the user may provide an Jacobian function `eqsdiff_jac!` to
+evaluate in-place the Jacobian; the user-defined Jacobian function must have the
+call signature `eqsdiff_jac!(jac, t, x, dx) -> nothing`. Otherwise, the current
+value of the Jacobian is computed via automatic differentiation using
+`TaylorSeries.jl`.
 
 """
-function liap_jetcoeffs!(eqsdiff!, t::Taylor1{T}, x::Vector{Taylor1{U}},
+function lyap_jetcoeffs!(eqsdiff!, t::Taylor1{T}, x::Vector{Taylor1{U}},
```
lbenet: What about exploiting multiple dispatch by defining two methods, one with and the other without `eqsdiff_jac!`. This would avoid the `if`!

lbenet: Another option is to have *always* the provided jacobian function, with `(jac, dδx)->jacobian!(jac, dδx)` as the default for

src/lyapunovspectrum.jl
```diff
@@ -157,18 +182,22 @@ function liap_taylorstep!(f, t::Taylor1{T}, x::Vector{Taylor1{U}}, dx::Vector{Ta
 end
 
 """
-    liap_taylorinteg(f, q0, t0, tmax, order, abstol; maxsteps::Int=500)
+    lyap_taylorinteg(f!, q0, t0, tmax, order, abstol[, f!]; maxsteps::Int=500)
```
lbenet: Probable some sort of typo here.

src/lyapunovspectrum.jl
```diff
@@ -229,17 +259,15 @@ function liap_taylorinteg(f, q0::Array{U,1}, t0::T, tmax::T,
...
             λtsum[ind] += log(RH[ind,ind])
             λ[ind,nsteps] = λtsum[ind]/tspan
         end
         for ind in eachindex(QH)
             @inbounds x0[dof+ind] = QH[ind]
         end
-        for i in eachindex(x0)
-            @inbounds x[i] = Taylor1( x0[i], order )
-        end
+        @inbounds x .= Taylor1.( x0, order )
```
lbenet: I am not sure if `@inbounds` has any effect here, or not.

src/lyapunovspectrum.jl
```diff
@@ -157,18 +182,22 @@ function liap_taylorstep!(f, t::Taylor1{T}, x::Vector{Taylor1{U}}, dx::Vector{Ta
...
+`lyap_taylorinteg` throws an `AssertionError`.  Optionally, the user may provide
+an Jacobian function `f!` to evaluate the current value of the Jacobian.
+Otherwise, the current value of the Jacobian is computed via automatic
+differentiation using `TaylorSeries.jl`. For more details on `f!`, see
+[`lyap_jetcoeffs!`](@ref).
 
 """
-function liap_taylorinteg(f, q0::Array{U,1}, t0::T, tmax::T,
-        order::Int, abstol::T; maxsteps::Int=500) where {T<:Real, U<:Number}
+function lyap_taylorinteg(f!, q0::Array{U,1}, t0::T, tmax::T,
```
lbenet: Same comment as above about changing `Nothing` for some function that reflects the current behavior.

lbenet: Also, instead of calling this optional parameter `df!`, I think something like `jacobian` makes clearer its role.

src/lyapunovspectrum.jl
PerezHz: > What about exploiting multiple dispatch by defining two methods, one with and the other without eqsdiff_jac!. This would avoid the if!

Thanks for the suggestion, actually that was the way I did things in the first place, but then I thought this way of doing things would avoid duplicated code, and after doing some benchmarks I saw there was no performance regression

> Another option is to have always the provided jacobian function, with (jac, dδx)->jacobian!(jac, dδx) as the default for...

Thanks, I will try this!

lbenet:
I like the present approach much more, and I still think there is room to optimize it here, and maybe even more with `@taylorize` or something similar. This is really nice and precisely opens up a new possibility.

The `if` block has been moved to `lyap_taylorstep!`. When `jac!` is provided, two calls to `jetcoeffs!` are made. The first one to obtain the `Taylor1` polynomials for `x`. The Jacobian matrix `jac` is updated (in place) the matrix through the provided function, which I'll write `jac!(t, x, jac)`. Then, the second call to `jetcoeffs!` is made. (Note that the calculation of the Jacobian may depends on `t` and `x`, but not on `dx`, which is why I don't have it there.) 

If `jac!` is not provided, currently a call to `lyap_jetcoeffs!` is made, which I think can be optimized. My suggestion is to make the first call to `jetcoeffs!` as above, which involves only `Taylor1` polynomials. We can then construct the Jacobian as we [currently do](https://github.com/PerezHz/TaylorIntegration.jl/pull/55/files/4723e4882ba5c21cbfead14a809bcc600385c334..c90499da78a80ab983a7ff372558c2584731f930#diff-370c3985beb25e9293dc0bde953ded11R127); note that here is where we need to use `_δv` which allow us to obtain the Jacobian through automatic differentiation. We can then make the *same* second call to `jetcoeffs!`, as we do it when `jac!` is provided. 

In other words, the `if` block should only affect the way `jac` is constructed; the rest is essentially identical. Note that if this idea works, we need to make *only one* call which involves `TaylorN` variables of the equations of motion (differential equations), instead of one for each calculated coefficient of the `Taylor1` polynomials.

src/lyapunovspectrum.jl
PerezHz: Thanks for noticing this! the `[...]` were intended to denote an optional parameter, but perhaps I'm using the wrong convention for optional arguments in the docstrings? Anyway, this is still WIP so the final signature call may change

src/lyapunovspectrum.jl
PerezHz: > Same comment as above about changing `Nothing` for some function that reflects the current behavior.

Thanks for the suggestion, and I agree that might be better! I'll work on this once agree on a final implementation for the changes introduced on this PR

> Also, instead of calling this optional parameter df!, I think something like jacobian makes clearer its role.

Sure, thanks! I'll change that

PerezHz:
Thanks a lot @lbenet for all your comments and suggestions! Commit a16a43d35586b27968242271b142f5d187a3c921, while getting closer to the desired design, presented a performance regression on the user-provided Jacobian case. Commit 4c4593eb767457cdb6d27a427e5a8bbdffa3609a is a provisional solution to handle both cases without loss of performance

lbenet:
I've done some benchmarks, using the scattered `@time` that are left in `test/lyapunov.jl`. This is a very brief summary:
 - Since a16a43d, the timings are essentially the same when the user provides a function to compute the Jacobian.
- Regarding the timings when computing the Lyapunov spectrum with AD, since a16a43d all timings are similar, which are are a factor 2 speed-up with respect to master; not bad! The best timings correspond to 4c4593e, which are about 10-15% faster than the others, in particular the last two.

I am not quite sure why 4c4593e is the best, but might be worth checking it.

lbenet:
Maybe my micro-benchmarks were too micro; forget then the previous comment (unless you can reproduce it).

src/lyapunovspectrum.jl
```diff
@@ -87,24 +87,25 @@ function modifiedGS!(A, Q, R, aⱼ, qᵢ, vⱼ)
...
-and `xaux`, `δx` and `dδx` are auxiliary vectors.
+Similar to [`jetcoeffs!`](@ref) for the calculation of the Lyapunov spectrum.
+`jac` is the current value of the linearization of the equations of motion
+(i.e., the Jacobian), and `xaux`, `δx`, `dδx` and `_δv` are auxiliary vectors.
+The current value of the Jacobian is computed via automatic differentiation using
+`TaylorSeries.jl`.
 
 """
-function liap_jetcoeffs!(eqsdiff!, t::Taylor1{T}, x::Vector{Taylor1{U}},
+function lyap_jetcoeffs!(eqsdiff!, t::Taylor1{T}, x::Vector{Taylor1{U}},
```
lbenet: I think now `lyap_jetcoeffs!` is not used at all, so we should simply delete it. Note that some docstrings reference to this function, so they will have to be updated.

src/lyapunovspectrum.jl
```diff
@@ -131,44 +130,76 @@ function liap_jetcoeffs!(eqsdiff!, t::Taylor1{T}, x::Vector{Taylor1{U}},
...
-spectrum. `jac` is the linearization of the equations of motion,
-and `xaux`, `δx`, `dδx` and `vT` are auxiliary vectors.
+function variational_eqs!(t::Taylor1{T}, x::Vector{Taylor1{U}},
+        dx::Vector{Taylor1{U}}, jac::Array{Taylor1{U},2}, eqsdiff_jac! =Nothing) where {T<:Real, U<:Number}
+    # Dimensions of phase-space: dof
+    nx = length(x)
+    dof = size(jac, 1)
+    if eqsdiff_jac! != Nothing
+        # Stability matrix
+        eqsdiff_jac!(jac, t, x)
```
lbenet: I most admit that I dislike the name `eqsdiff_jac!`; it reminds me to be some sort of differential equation related to the jacobian (which are the variationals!), but refers to the user provided calculation of the Jacobian. I think that `userjacobian!`, `jacobianfunc!` or something like that (with or without `_`) is descriptive enough to be a good name.

src/lyapunovspectrum.jl
```diff
@@ -131,44 +130,76 @@ function liap_jetcoeffs!(eqsdiff!, t::Taylor1{T}, x::Vector{Taylor1{U}},
...
+        dx::Vector{Taylor1{U}}, xaux::Vector{Taylor1{U}},
+        δx::Array{TaylorN{Taylor1{U}},1}, dδx::Array{TaylorN{Taylor1{U}},1},
+        jac::Array{Taylor1{U},2}, t0::T, t1::T, x0::Array{U,1}, order::Int,
+        abstol::T, _δv::Array{TaylorN{Taylor1{U}}}, jac! =Nothing) where {T<:Real, U<:Number}
+    # Dimensions of phase-space: dof
+    nx = length(x)
+    dof = length(δx)
+    jetcoeffs!(f!, t, view(x, 1:dof), view(dx, 1:dof), view(xaux, 1:dof))
+    if jac! == Nothing
+        # lyap_jetcoeffs!(f!, t, x, dx, xaux, δx, dδx, jac, _δv)
```
lbenet: I think the contents of the first part of the `if` (at least partially) could be moved to a function. I am not sure if that would be better for performance, but it would allow testing the user-provided function for the Jacobian against AD, and that's time saving!

src/lyapunovspectrum.jl
lbenet: Regarding this, I was simply confused; perhaps, it was the "," in between: `[, ...]`, instead of `, [...]`.

src/lyapunovspectrum.jl
lbenet: See comment above about the name of that function.

test/lyapunov.jl
```diff
@@ -7,14 +7,28 @@ using Test
...
+#Lorenz system ODE:
+function lorenz!(t, x, dx)
+    dx[1] = σ*(x[2]-x[1])
+    dx[2] = x[1]*(ρ-x[3])-x[2]
+    dx[3] = x[1]*x[2]-β*x[3]
+    nothing
+end
+
+#Lorenz system Jacobian (in-place):
+function lorenz_jac!(jac, t, x)
```
lbenet: In the following lines, zero(x[1]) is preferred over 0.0*x[1], though this should be tested.

lbenet:
Merged. Thanks a lot!

PR #56, opened 2018-09-27
Further performance improvement of `lyap_taylorinteg`: `lyap_jetcoeffs!` strikes back!

PerezHz:
This PR is aimed towards improving performance of `lyap_taylorinteg`. The changes introduced by this PR are mainly the following:

- Compute the Jacobian `jac`, either by automatic differentiation or by the Jacobian function provided by the user, in a separate function `stabilitymatrix!`.
- Bring back `lyap_jetcoeffs!`, which now computes *only* the jet of coefficients in the vector of dependent variables `x` associated  to the variational equations. Another difference is that instead of passing the differential equations `eqsdiff!` as an argument to this function, only the Jacobian `jac` is passed. The implementation of `lyap_jetcoeffs!` was actually obtained meta-programatically using the `@taylorize` macro (see #31 ), with some manual tweaks.
- Tests and docstrings have been updated accordingly.

### DIFF
--- a/src/lyapunovspectrum.jl
+++ b/src/lyapunovspectrum.jl
@@ -113,40 +113,43 @@ function lyap_jetcoeffs!(t::Taylor1{T}, x::AbstractVector{Taylor1{S}},
         dx::AbstractVector{Taylor1{S}}, jac::Matrix{Taylor1{S}},
         varsaux::Array{Taylor1{S},3}) where {T <: Real, S <: Number}
     order = t.order
-    # number of Lyapunov exponents
-    nlyaps = size(jac, 1)
-    inds = axes(jac, 1)
-    # 0-th order evaluation of variational equations
+    # `dofrange` behaves like 1:dof, where `dof = size(jac, 1)`. Used to initialize `b` and `I`
+    dofrange = axes(jac, 1)
+    # The `view` below allows us to obtain CartesianIndices from `varsaux` when calling `eachindex(b)`
+    # `varsaux` is an array with size n×n×n
+    b = view(varsaux, dofrange, dofrange, dofrange)
+    # We use `I` to map between cartesian and linear indices
+    I = LinearIndices((dofrange, dofrange))
+
+    # 0-th order evaluation of variational equations `dx = jac * x`
+    # Initialize LHS of variational equations at zero
     dx .= zero(x[1])
-    for j in inds
-        for i in inds
-            for k in inds
-                varsaux[k, i, j] = Taylor1(constant_term(jac[i, k]) * constant_term(x[nlyaps * (j - 1) + k]), order)
-                dx[nlyaps * (j - 1) + i] = Taylor1(constant_term(dx[nlyaps * (j - 1) + i]) + constant_term(varsaux[k, i, j]), order)
-            end
-        end
+    # Compute 0-th Taylor coefficients of matrix product `jac * x` and save into `dx`
+    for i in eachindex(b)
+        varsaux[i] = Taylor1(constant_term(jac[ i[1], i[3] ]) * constant_term(x[ I[i[3], i[2]] ]), order)
+        lin_indx = I[ i[1], i[2]] # map from cartesian to linear index
+        dx[lin_indx] = Taylor1(constant_term(dx[lin_indx]) + constant_term(varsaux[i]), order)
     end
     # Recursion relations, 0-th order
     for __idx = eachindex(x)
         (x[__idx]).coeffs[2] = (dx[__idx]).coeffs[1]
     end
-    # Compute Taylor coefficients of variational equations up to order `order`
+
+    # Compute Taylor coefficients of variational equations `dx = jac * x` up to order `order`
     for ord = 1:order - 1
         ordnext = ord + 1
-        for j in inds
-            for i in inds
-                TaylorSeries.zero!(dx[nlyaps * (j - 1) + i], x[1], ord)
-                for k in inds
-                    TaylorSeries.mul!(varsaux[k, i, j], jac[i, k], x[nlyaps * (j - 1) + k], ord)
-                    TaylorSeries.add!(dx[nlyaps * (j - 1) + i], dx[nlyaps * (j - 1) + i], varsaux[k, i, j], ord)
-                end
-            end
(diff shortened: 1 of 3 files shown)


lbenet:
Have you checked if this leads PR to performance improvements?

src/lyapunovspectrum.jl
```diff
@@ -86,16 +96,54 @@ function modifiedGS!(A, Q, R, aⱼ, qᵢ, vⱼ)
...
+
+"""
+function lyap_jetcoeffs!(t::Taylor1{T}, x::AbstractVector{Taylor1{S}},
+        dx::AbstractVector{Taylor1{S}}, jac::Matrix{Taylor1{S}}) where {T <: Real, S <: Number}
+    order = t.order
+    nlyaps = size(jac, 1) # number of Lyapunov exponents
+    temp_002 = Array{eltype(x)}(undef, nlyaps, nlyaps, nlyaps)
+    temp_001 = Array{Taylor1{S}}(undef, size(temp_002))
+    @__dot__ temp_001 = Taylor1(zero(S), order)
+    for j = 1:nlyaps
```
lbenet: If I understand correctly the algorithm, it actually performs explicitly the matrix multiplication to integrate the variationals, with a lot of care in the indices.  What about assuming that the indices of `x` and `dx` begin at 1? In that case, `lyap_jetcoeffs!` would be called via a view or something similar? My point is that such a change would make simpler the indexing, actually saving some operations.

lbenet:
Thanks a lot for the PR. I guess that making explicit the matrix multiplication (using low level functions) can certainly lead to performance improvements. I think we should simply assume a nicer form of the entries, to simplify the notation (indexing).

PerezHz:
> Have you checked if this leads PR to performance improvements?

Yes, there are performance improvements!

For example, see below some tests of `test/lyapunov.jl`. The first, third and fifth `@time` outputs correspond to integrations with `jacobianfunc! == Nothing`; i.e., the Jacobian is computed by autodiff. The second, fourth and sixth `@time` outputs correspond to integrations with `jacobianfunc!` provided by the user. On my machine in julia 1.0 with current TaylorIntegration master:
screen shot 2018-09-27 at 10 40 58
And with #56:
screen shot 2018-09-27 at 10 41 14
Tests in travis and appveyor show similar trends

lbenet:
Thanks @PerezHz , the improvement is quite clear! Have you consider my suggestion for reeindexing the arrays which are used in `lyap_jetcoeffs!`?

src/lyapunovspectrum.jl
PerezHz: Yes, exactly, the matrix multiplication was implemented explicitly in order to avoid the use of `reshape` and also to perform operations in-place. About the indexing, `x` and `dx` are already assumed to begin at 1 and called with views over `x` and `dx` (see, for example, `lyap_jetcoeffs!` call in `lyap_taylorstep!`, line 174 of `src/lyapunovspectrum.jl`). The ugly indexing (with `1:nlyaps`, etc.) comes from trying to implement the matrix multiplication of, say, `dx = jac*x` where `jac` is a square matrix with size `(nlyaps, nlyaps)` but `x` and `dx` are actually 1-dim arrays (i.e, vectors) with size `nlyaps*nlyaps`

PerezHz:
> Thanks @PerezHz , the improvement is quite clear! Have you consider my suggestion for reeindexing the arrays which are used in lyap_jetcoeffs!?

Just commented above something related to that; thank you for all your comments and suggestions!

PerezHz:
Figured out a way to avoid use of a temporary variable in `lyap_jetcoeffs!` and got rid of it; performance improved even more!

src/lyapunovspectrum.jl
lbenet: Thanks for clarifying this; perhaps you should add some comments in the program to remind us.

I would then suggest to define the abbreviations (change the names at your will):
```jl
indx1 = nlyaps * (j - 1) + i  # after line 119, and also after 134 
indx2 = nlyaps * (j - 1) + k  # after line 121, and also after 136
```
and use them in the lines following those definitions. This may lead to some insignificant performance improvement, and clarity.

lbenet:
@PerezHz Check [these suggestions](https://github.com/PerezHz/TaylorIntegration.jl/pull/56#discussion_r220995555). Ping me when this is ready for review.

PerezHz:
> @PerezHz Check these suggestions. Ping me when this is ready for review.

Than you @lbenet, I've tried to address your suggestions with the last commit; this is ready now for review

PerezHz:
That is, unless travis and appveyor think otherwise... but they're almost all green by now

PerezHz:
Green lights!

src/lyapunovspectrum.jl
```diff
@@ -86,51 +96,89 @@ function modifiedGS!(A, Q, R, aⱼ, qᵢ, vⱼ)
...
+        dx::AbstractVector{Taylor1{S}}, jac::Matrix{Taylor1{S}},
+        varsaux::Array{Taylor1{S},3}) where {T <: Real, S <: Number}
+    order = t.order
+    # `dofrange` behaves like 1:dof, where `dof = size(jac, 1)`. Used to initialize `b` and `I`
+    dofrange = axes(jac, 1)
+    # The `view` below allows us to obtain CartesianIndices from `varsaux` when calling `eachindex(b)`
+    # `varsaux` is an array with size n×n×n
+    b = view(varsaux, dofrange, dofrange, dofrange)
+    # We use `I` to map between cartesian and linear indices
+    I = LinearIndices((dofrange, dofrange))
```
lbenet: I think `I` is not the best name, since `LinearAlgebra` exports `I` which is the identity matrix; I thus suggest to change the name of that variable.

test/REQUIRE
```diff
@@ -1,2 +1,2 @@
-TaylorSeries 0.7.4
-Elliptic 0.5.0
+TaylorSeries v0.8.0
```
lbenet: Good catch!

lbenet:
Excellent! Merging!

PerezHz:
Great, thanks for merging! Should we re-rebase now #31?

PR #60, opened 2018-10-28
Save gif-files in docs/build

### LINKED ISSUE #59: Some gif's are not displayed in the docs
Opening this issue to remember that the gifs in the docs for [Poincaré maps](https://perezhz.github.io/TaylorIntegration.jl/latest/root_finding.html) are not displayed.

lbenet:
This may fix #59. Locally it works, but I am not sure it does here.

### DIFF
--- a/docs/src/root_finding.md
+++ b/docs/src/root_finding.md
@@ -138,7 +138,7 @@ poincare_anim1 = @animate for i=1:21
     ylabel!("py")
     title!("Hénon-Heiles Poincaré map (21 iterates)")
 end
-gif(poincare_anim1, "henonheilespoincaremap5.gif", fps = 2)
+gif(poincare_anim1, "../build/henonheilespoincaremap5.gif", fps = 2)
 nothing # hide
 ```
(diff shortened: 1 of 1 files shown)


PerezHz:
Thanks for this PR!

We could try with this and see if it just works! Otherwise, from what I've seen in the Plots.jl docs (see e.g. [here](https://github.com/JuliaPlots/PlotDocs.jl/blob/master/docs/src/index.md#simple-is-beautiful)), they don't actually let travis build the gif files; instead, they upload the gif files as part of a repo and then just post them where it corresponds, so I guess that would be another option just in case another option is needed.

lbenet:
I'm in favor for testing what happens when merging. Locally, the file is (properly) saved in the `builds/` directory, which I think is the folder that gets uploaded. If this doesn't work, we can revert it. 

But we should think it over before doing anything.

PerezHz:
Just merged to see what happens; as you said, if this doesn't work, then we may revert it

lbenet:
It seems it didn't work...

lbenet:
We should ask in discourse, or open an issue in Documenter.jl

PR #62, opened 2019-01-17
Add taylorinteg root-finding method with time ranges

PerezHz:
This PR adds a method with root-finding capabilities to `taylorinteg`, with the following call signature:

```julia
taylorinteg(f!, g, q0::Array{U,1}, trange::AbstractVector{T},
        order::Int, abstol::T; maxsteps::Int=500, parse_eqs::Bool=true,
        eventorder::Int=0, newtoniter::Int=10, nrabstol::T=eps(T)) where {T <: Real,U <: Number}
```

That is, the new method allows to find roots of the scalar function `g`, while evaluating the solution only at times specified by `trange`.

TODO: add tests, docs

EDIT: This PR also optimizes the internal handling of `taylorinteg` methods with time ranges.


lbenet:
Thanks for the contribution! I'll try to take a look on it later. Yet, I'd like to point out that there are some tests failing in appveyor, while there is no problem in travis. Any ideas?

PerezHz:
> Thanks for the contribution! I'll try to take a look on it later. Yet, I'd like to point out that there are some tests failing in appveyor, while there is no problem in travis. Any ideas?

Thanks for the heads up! The error was related to too stringent tolerances in a couple of tests; I fixed it for the time being by relaxing the tolerances a bit

src/explicitode.jl
```diff
@@ -427,7 +427,7 @@ Note that f! updates (mutates) the pre-allocated vector dx.
 Note that the initial conditions `q0TN` are of type `TaylorN{Float64}`.
 
 """
-function taylorinteg(f, x0::U, trange::Union{AbstractRange{T},Vector{T}},
+function taylorinteg(f, x0::U, trange::AbstractVector{T},
```
lbenet: Nice catch!

src/rootfinding.jl
```diff
@@ -206,3 +206,145 @@ function taylorinteg(f!, g, q0::Array{U,1}, t0::T, tmax::T,
...
+    taylorinteg(f, g, x0, trange, order, abstol; kwargs... )
+
+Root-finding method of `taylorinteg`; returns the solution of the ODE evaluated
+*only* at the times specified by `trange`. Given a function `g(t, x, dx)`,
+called the event function, `taylorinteg` checks for the occurrence of a root
+of `g` evaluated at the solution; that is, it checks for the occurrence of an
+event or condition specified by `g=0`. Then, `taylorinteg` attempts to find that
+root (or event, or crossing) by performing a Newton-Raphson process. When
+called with the `eventorder=n` keyword argument, `taylorinteg` searches for the
+roots of the `n`-th derivative of `g`, which is computed via automatic
```
lbenet: Question: I guess that `eventorder` should satisfy `order >= eventorder`, i.e., it should be smaller than the order of series expansion used to approximate the solution. Is that checked somewhere?

PerezHz: Yes, `eventorder` should satisfy `order >= eventorder` but currently that is not checked, I will add an assertion, many thanks for noticing this!

PerezHz: Just added some `@asserts` which check this condition, thanks again!

PerezHz:
> I still think that the function g (used to define the "surface of section") should be better explained in the docstrings and in the docs. Somehow the explanation confuses me, in particular with respect to the way it is used in the examples in the documentation. I think it is worth to clarify all that in this PR.

I agree! Will push a commit asap with updated and better-explained docs!

PerezHz:
Tests are passing in travis and appveyor, but it seems that coverage services are broken (see JuliaCI/Coverage.jl#187), since lines that are covered by the tests are not being marked as covered by neither coveralls nor codecov.

src/rootfinding.jl
lbenet: Thank's a lot, I think this is solved.

test/lyapunov.jl
```diff
@@ -175,51 +173,55 @@ end
...
     @test lorenztr == -(1+σ+β)
 
-    xw, λw = lyap_taylorinteg(lorenz!, q0, t0:1.0:tmax, _order, _abstol; maxsteps=2)
+    trange = t0:1.0:tmax
+    xw, λw = lyap_taylorinteg(lorenz!, q0, trange, _order, _abstol; maxsteps=5)
     @test xw[1,:] == q0
-    @test size(xw) == (length(t0:1.0:tmax), 3)
-    @test size(λw) == (length(t0:1.0:tmax), 3)
+    @test size(xw) == (length(trange), 3)
+    @test size(λw) == (5+1, 3)
```
lbenet: Why not writing this test as `@test size(λw) == (6, 3)`, and add a comment (at the end of the line?) that 6 = maxsteps+1.

test/lyapunov.jl
```diff
@@ -175,51 +173,55 @@ end
...
-    xw_, λw_ = lyap_taylorinteg(lorenz!, q0, t0:1.0:tmax, _order, _abstol, lorenz_jac!; maxsteps=2)
+    xw2, λw2 = lyap_taylorinteg(lorenz!, q0, collect(trange), _order, _abstol; maxsteps=5)
+    @test xw[1, :] == xw2[1, :]
+    @test λw == λw2
+    xw_, λw_ = lyap_taylorinteg(lorenz!, q0, trange, _order, _abstol, lorenz_jac!; maxsteps=5)
     @test xw[1,:] == q0
-    @test size(xw) == (length(t0:1.0:tmax), 3)
-    @test size(λw) == (length(t0:1.0:tmax), 3)
+    @test size(xw) == (length(trange), 3)
+    @test size(λw) == (5+1, 3)
```
lbenet: Same as above...

test/lyapunov.jl
```diff
@@ -175,51 +173,55 @@ end
...
+    trange = t0:1.0:tmax
+    xw, λw = lyap_taylorinteg(lorenz!, q0, trange, _order, _abstol; maxsteps=5)
     @test xw[1,:] == q0
-    @test size(xw) == (length(t0:1.0:tmax), 3)
-    @test size(λw) == (length(t0:1.0:tmax), 3)
+    @test size(xw) == (length(trange), 3)
+    @test size(λw) == (5+1, 3)
     @test prod(isnan.(xw[2:end,:]))
-    xw_, λw_ = lyap_taylorinteg(lorenz!, q0, t0:1.0:tmax, _order, _abstol, lorenz_jac!; maxsteps=2)
+    xw2, λw2 = lyap_taylorinteg(lorenz!, q0, collect(trange), _order, _abstol; maxsteps=5)
```
lbenet: I just performed some benchmarks, and it seems better (in terms of time and allocations) to use `vec(trange)` instead of `collect(trange)`. This may also be useful elsewhere in our code; the important thing to keep in mind is that `vec` reshapes to a vector, but shares the memory, so changing the "new" vector also changes the old array.

lbenet: This also applies somewhere below...

test/lyapunov.jl
PerezHz: Thanks, I just addressed both lines

test/lyapunov.jl
PerezHz: Done, thanks!

lbenet:
> Tests are passing in travis and appveyor, but it seems that coverage services are broken (see JuliaCI/Coverage.jl#187), since lines that are covered by the tests are not being marked as covered by neither coveralls nor codecov.

I think we shouldn't care too much about those problems, for the time being, at least.

However, i am a bit more concerned about #31. To avoid another complicate rebase in #31, I would be in favor of merging that PR before this one.

test/lyapunov.jl
PerezHz: Just to be sure how to proceed, should I try to replace the occurrences of `collect` in the code with `vec`?

PerezHz:
> However, i am a bit more concerned about #31. To avoid another complicate rebase in #31, I would be in favor of merging that PR before this one.

Ok, I agree! I will check #31 right away!

test/lyapunov.jl
lbenet: I am not quite sure, but at least it is worth testing. In the case marked (and other similar) there is no risk that `trange` be changed, so I would simply test what happens if this is changed. Other occurrences of `collect` should be carefully addressed.

lbenet:
#31 is merged, which induced some conflicts. Can you rebase to current master?

PerezHz:
> #31 is merged, which induced some conflicts. Can you rebase to current master?

Just rebased to current master and fixed conflicts, tests should pass but if there's anything I missed I'll fix it

test/lyapunov.jl
PerezHz: Actually almost all occurrences of collect were in this same file (test/lyapunov.jl), so I went ahead and changed it

lbenet:
LGTM. Updating the documentation is what is still to be done.

lbenet:
LGTM. Is this ready to be merged?

PerezHz:
> LGTM. Is this ready to be merged?

Great! Yes, it's ready to be merged!

lbenet:
Merging...

PerezHz:
Thanks!

PR #64, opened 2019-01-23
Fix Lyapunov spectrum integrator with time ranges/vectors

PerezHz:
After merging #62 it was noted that, while orbits and Lyapunov spectra computed with `lyap_taylorinteg` are now independent of the time range/vector provided by the user, the Lyapunov exponents were not being evaluated at each point specified by the time range/vector. This PR fixes this issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ os:
 julia:
   - 0.7
   - 1.0
+  - 1.1
   - nightly
 
 notifications:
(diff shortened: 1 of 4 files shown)


PerezHz:
Locally, tests pass; if everything continues this way, I'll just wait for the green lights before merging and then v0.4.0 will be tagged

lbenet:
Thanks a lot! Looking very much forward to v0.4.0

PerezHz:
> Incidentally, maybe this is a good opportunity to add testing Julia 1.1 in travis and appveyor!

Yes, I agree! I'll push that change right away!

PerezHz:
Merging!

lbenet:
Thanks a lot!

Issue #69, opened 2019-04-24
Align dynamics function arguments to DifferentialEquations API


jessebett:
Currently `jetcoeffs!(eqsdiff!::Function, t, x, dx, xaux)` expects `eqsdiff!` to follow with `taylorinteg` so `f!(t, x, dx)`. However, it would be much cleaner if this followed the updated DifferentialEquations.jl convention, where in-place dynamics equations are given as `f!(dx,x,p,t)` where `p` are parameters for `f!`. 

e.g. https://docs.juliadiffeq.org/stable/tutorials/ode_example.html#Example-2:-Solving-Systems-of-Equations-1

lbenet:
Thanks for the suggestion! 

Some time ago, we discussed this with @PerezHz, and it is certainly something to consider. While the core of `TaylorIntegration` would require minor modifications (at least for the "typical cases"), it is not clear to me if `@taylorize` would be so easily handled. `@taylorize` is important because it allows to construct more efficient (internal) implementations of `jetcoeffs!` for the specific `f!`.

Please give us a bit of time to think this over and play around!

lbenet:
See #70

lbenet:
Closing, since #70 has been merged and v0.5.0 has been tagged.

PR #70, opened 2019-05-27
Use DiffEqs convention to define the ODEs function

lbenet:
This is work in progress addressing #69.

Aside from the convenience of using the DiffEqs convention to define the differential equations, I think this approach allows to exploit (at least in some cases) the macro `@taylorize` from the DiffEqs common interface, thus allowing to get better performance. The cases that I think work are those in which the arguments `du` and `u` (in `f(du, u, p, t)`, where `f` is the function that defines the differential equations) are vectors; I don't think more generic forms allow the use of the macro, due to some assumptions in the form of the macro.

I have also being able to extend it a little bit, which makes certain things easier.

I'll be working updating the docs, but this will go slower due to time constraints. 

This is certainly a major breaking change, so we should eventually tag a new version, and make it as clear as possible for the user.

Any comments are really appreciated.

cc @jessebett @ChrisRackauckas


src/common.jl
```diff
@@ -40,29 +41,32 @@ function DiffEqBase.solve(
...
+        f! = (du, u, p, t) -> (du .= vec(f(reshape(u, sizeu), p, t)); 0)
+        t, vectimeseries = taylorinteg(f!, prob.u0, prob.tspan[1], prob.tspan[2],
+            alg.order, abstol, prob.p, maxsteps=maxiters)
 
-    t,vectimeseries = taylorinteg(f!, u0, prob.tspan[1], prob.tspan[2], alg.order,
-                                                      abstol, maxsteps=maxiters)
+    elseif isinplace && typeof(prob.u0) <: AbstractArray{eltype(uType),2}
+        f! = (du, u, p, t) -> (
+                dd = reshape(du, sizeu); uu = reshape(u, sizeu);
+                f(dd, uu, p, t); u .= vec(uu); du .= vec(dd); 0)
```
ChrisRackauckas: You don't need to `u .= vec(uu); du .= vec(dd)` since a `reshape` is a view.

lbenet: Thanks Chris! I'll change it

src/common.jl
```diff
@@ -40,29 +41,32 @@ function DiffEqBase.solve(
...
-    elseif !isinplace && typeof(prob.u0)<:AbstractArray
-        f! = (t, u, du) -> (du .= vec(f(reshape(u, sizeu), p, t)); 0)
-    elseif typeof(prob.u0)<:Vector{Float64}
-        f! = (t, u, du) -> f(du, u, p, t)
-    else # Then it's an in-place function on an abstract array
-        f! = (t, u, du) -> (f(reshape(du, sizeu), reshape(u, sizeu), p, t);
-                            u = vec(u); du=vec(du); 0)
-    end
+    if !isinplace && typeof(prob.u0) <: Vector{Float64}
+        f! = (du, u, p, t) -> (du .= f(u, p, t); 0)
```
lbenet: @ChrisRackauckas I guess the same applies here, right?

ChrisRackauckas: No, this will need to `.=`.

lbenet: I just found it out! You were faster!

ChrisRackauckas:
Looks awesome. It would be interesting to see how this does on @sebastian-mc's benchmarks https://nbviewer.jupyter.org/github/JuliaDiffEq/DiffEqBenchmarks.jl/blob/master/DynamicalODE/Henon-Heiles_energy_conservation_benchmark.ipynb (updated in https://github.com/JuliaDiffEq/DiffEqBenchmarks.jl/pull/32).

lbenet:
Thanks for the suggestion; I'll take a look on that later, as time allows me...

lbenet:
@ChrisRackauckas I've checked it, and under certain circumstances, using `@taylorize` also improves the performance when called from DiffsEqs.

ChrisRackauckas:
>using @taylorize also improves the performance when called from DiffsEqs.

How much of a speedup? I assume you just throw it on the `f` definition and it defines a bunch of new dispatches?

lbenet:
For the pendulum is a factor ~3-4; I have other examples which is a factor ~8. The more involved the example, better is the speedup.

ChrisRackauckas:
That sounds like a good amount. Probably happens even more at lower tolerances. We'll get the benchmarks updated to use `@taylorize`.

lbenet:
Using `@taylorize` is a bit subtle and may require a bit extra work, e.g., to make sure that the AST reduces to unary or binary operations, though this may not so important for Hénon-Heiles.

Incidentally, I did some benchmarks in the restricted circular three-body problem, comparing `Vern9()`, `Feagin14()` and `TaylorMethod()`. With `abstol = 1e-20` only, `TaylorMethod(25)` was terribly slow in comparison with the others, but the others were so inaccurate (with respect to the constant of motion) that the results were simply meaningless. Then I imposed `abstol = 1e-20` and `reltol=1e-20` to attain similar accuracy; then `TaylorMethod()` together with `@taylorize` was faster by a factor 2 against `Feagin14()`; I couldn't get the results for `Vern9()`. I will clean-up that notebook and gist it, so you can take a look on it.

lbenet:
@ChrisRackauckas Two questions:
- Which lower version should we use for `DiffEqsBase` (now in REQUIRE, but eventually in Project.toml)? Currently we require at least v4.21.3
- This will be a breaking change, so a new tag (minor version at least) seems reasonable. I recall that there was a warning/info message displayed quite clearly when updating `DiffEqs` when these sort of things happened. Can you point me to the file where this was done?

Thanks a lot for your comments in this PR!

ChrisRackauckas:
>Which lower version should we use for DiffEqsBase (now in REQUIRE, but eventually in Project.toml)? Currently we require at least v4.21.3

v5.0.0, since that's where the v1.0 compatibility came in.

>This will be a breaking change, so a new tag (minor version at least) seems reasonable. I recall that there was a warning/info message displayed quite clearly when updating DiffEqs when these sort of things happened. Can you point me to the file where this was done?

We just put print commands in the deps/build.jl file.

lbenet:
The tests that aren't passing are for Julia v0.7; obviously related to requiring v5.0.0 of DiffEqBase...

@PerezHz How about dropping Julia v0.7 compatibility?

PerezHz:
> @PerezHz How about dropping Julia v0.7 compatibility?

I agree!

lbenet:
I think this is ready for reviewing. While working on this PR I noticed a few spot which I think can be improved. They imply a small redesign of some routines, but I maybe they will slightly improve  performance.

I will prepare the road to tag a new (minor) version, once this is merged.

lbenet:
Rebased to master, and corrected a problem that came out by trying to increase coverage

src/rootfinding.jl
```diff
@@ -109,51 +113,50 @@ called with the `eventorder=n` keyword argument, `taylorinteg` searches for the
...
-
-The current keyword arguments are `maxsteps=500`, `eventorder=0`,
-`newtoniter=10`, and `nrabstol=eps(T)`, where `T` is the common type of `t0`,
-`tmax` and `abstol`.
-
 The current keyword argument are:
-- `maxsteps=500`: maximum number of integration steps.
-- `parse_eqs=true`: usie the specialized method of `jetcoeffs!` created
+- `maxsteps[=500]`: maximum number of integration steps.
+- `parse_eqs[=true]`: usie the specialized method of `jetcoeffs!` created
```
PerezHz: *use

src/lyapunovspectrum.jl
```diff
@@ -189,9 +190,9 @@ function lyap_taylorstep!(f!, t::Taylor1{T}, x::Vector{Taylor1{U}},
     δt = stepsize(view(x, 1:dof), abstol)
     δt = min(δt, t1-t0)
 
-    # Update x0
-    evaluate!(x, δt, x0)
-
+    # # Update x0
```
PerezHz: Should this commented lines be removed?

src/rootfinding.jl
lbenet: Good catch! I'll change it

src/lyapunovspectrum.jl
lbenet: Thanks! Changed

test/many_ode.jl
```diff
@@ -8,27 +8,27 @@ const _order = 28
...
             Dx[i] = x[i]^2
         end
         nothing
     end
     exactsol(t, x0) = x0/(1.0-x0*t)
     t0 = 0.0
-    q0 = [3.0, 1.0]
-    q0T = [Taylor1(q0[1], _order), Taylor1(q0[2], _order)]
+    q0 = [3, 1]
+    q0T = [Taylor1(3.0, _order), Taylor1(1.0, _order)]
```
PerezHz: Maybe this could be simplified to `q0T = Taylor1.(q0, _order)`

docs/src/common.md
```diff
@@ -233,18 +247,28 @@ Finally, we comment on the time spent by each integration.
 ```@example common
 @elapsed solve(prob, Vern9(), abstol=1e-20);
 ```
-Clearly, the integration with `TaylorMethod()` takes *much longer* than that using
+The integration with `TaylorMethod()` takes *much longer* than that using
 `Vern9()`. Yet, as shown above, the former preserves the Jacobi constant
-to a high accuracy while displaying the correct dynamics; whereas the latter
-solution loses accuracy in the sense of not conserving the Jacobi constant,
+to a high accuracy, whereas the latter
+solution looses accuracy in the sense of not conserving the Jacobi constant,
```
PerezHz: Should it say "loses" with only one "o"?

src/rootfinding.jl
```diff
@@ -34,14 +34,14 @@ nrconvergencecriterion(g_val::U, nrabstol::T, nriter::Int,
         newtoniter::Int) where {U<:Number, T<:Real} = abs(constant_term(g_val)) > nrabstol && nriter ≤ newtoniter
 
 """
-    findroot!(g, t, x, dx, g_val_old, g_val, eventorder, tvS, xvS, gvS,
+    findroot!(t, x, dx, g_val_old, g_val, eventorder, tvS, xvS, gvS,
```
PerezHz: Great catch!

test/many_ode.jl
lbenet: Good point. I'll change it.

lbenet: Mmmm. Rereading the code I recall that the change has a purpose. `q0T` is passed to `jetcoeffs!`, so it should be a `Taylor1{Float64}` vector. I'll thus change it to `Taylor1.(Float64.(q0), _order)`

docs/src/common.md
lbenet: Good catch! Changed

src/rootfinding.jl
lbenet: :-)

lbenet:
I'll pushed the changes. I'll add the Project.toml and delete the REQUIRE files, so we can register a new version.

PerezHz:
> I'll pushed the changes. I'll add the Project.toml and delete the REQUIRE files, so we can register a new version.

Thanks! I just gave access to JuliaRegistrator and TagBot

lbenet:
Thanks for giving access to JuliaRegistrator and TagBot. 

I just pushed two more commits, which include a message (which appears when loading `TaylorIntegration` and in the `README.md`) which mention briefly the breaking changes.

@PerezHz Do you agree with what it is written? Should I extend it or shorten it? If you do, once all tests are green, I'll merge and register the new version.

PerezHz:
@lbenet I think it’s fine as the changes are now, and should any change be necessary we can modify it later

lbenet:
Merging

PR #75, opened 2019-06-13
Regression in stepsize

lbenet:
This solves a regression due to `stepsize`

### DIFF
--- a/src/explicitode.jl
+++ b/src/explicitode.jl
@@ -143,7 +143,7 @@ function stepsize(q::AbstractArray{Taylor1{U},1}, epsilon::T) where
 
     # If `isinf(h)==true`, we use the maximum (finite)
     # step-size obtained from all coefficients as above.
-    # Note that the time step is independent from `epsilon.
+    # Note that the time step is independent from `epsilon`.
     if isinf(h)
         h = zero(R)
         for i in eachindex(q)
--- a/src/explicitode.jl
+++ b/src/explicitode.jl
@@ -140,38 +140,39 @@ function stepsize(q::AbstractArray{Taylor1{U},1}, epsilon::T) where
         @inbounds hi = stepsize( q[i], epsilon )
         h = min( h, hi )
     end
+
+    # If `isinf(h)==true`, we use the maximum (finite)
+    # step-size obtained from all coefficients as above.
+    # Note that the time step is independent from `epsilon.
     if isinf(h)
         h = zero(R)
         for i in eachindex(q)
-            @inbounds hi = _finite_stepsize(q[i], epsilon)
+            @inbounds hi = _second_stepsize(q[i], epsilon)
             h = max( h, hi )
         end
     end
     return h::R
 end
 
-# Helper function to avoid code repetition
-function _stepsize(aux1::U, epsilon::T, k::Int) where {T<:Real, U<:Number}
+"""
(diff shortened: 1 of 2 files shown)


lbenet:

Tests seem to complete correctly, except for nightly (allowed to fail). Strangely, the same test is failing in travis and appveyor, but the numbers reported are different. See [here](https://travis-ci.org/PerezHz/TaylorIntegration.jl/jobs/545999010#L661) and [here](https://ci.appveyor.com/project/PerezHz/taylorintegration-jl/builds/25297948/job/jmh9sqf3m8dp7bne#L248).

src/explicitode.jl
```diff
@@ -112,48 +112,80 @@ Returns a maximum time-step for a the Taylor expansion `x`
...
+    h = convert(R, Inf)
     for i in eachindex(q)
         @inbounds hi = stepsize( q[i], epsilon )
         h = min( h, hi )
     end
-    return h
+
+    # If `isinf(h)==true`, we use the maximum (finite)
+    # step-size obtained from all coefficients as above.
+    # Note that the time step is independent from `epsilon.
```
PerezHz: `` `epsilon `` -> `` `epsilon` ``

src/explicitode.jl
```diff
@@ -193,6 +225,9 @@ function taylorstep!(f, t::Taylor1{T}, x::Taylor1{U}, t0::T, t1::T, order::Int,
 
     # Compute the step-size of the integration using `abstol`
     δt = stepsize(x, abstol)
+    if isinf(δt)
+        δt = _second_stepsize(x, abstol)
```
PerezHz: Just for consistency: is there any particular reason why `_second_stepsize` is called from `stepsize` in the vector case, but called from `taylorstep!` in the scalar case?

PerezHz:
@lbenet thanks for looking into this!

The error in nightly seems strange enough, but isn't things like this why we allow nightlies to fail in the first place? Otherwise, we could ask for less stringent tolerance in the failing test, e.g., `1e-14` instead of `1e-15`?

PerezHz:
LGTM, and I'm in favor of merging!

src/explicitode.jl
lbenet: Good catch, I'll push a new commit correcting this.

src/explicitode.jl
lbenet: Fair point; I guess is speed. In the vector case, `_second_stepsize` is called if the `min` of all the possible stepsizes is `Inf` (so only once), which I guess occurs seldom. In the scalar case, you have to check it consistently. Somewhere here was actually the problem that lead to the performance regression.

lbenet:
What bothers me about the error is that it depends on which OS you are running, and I have no idea where to look for such a problem.

I just pushed a new commit addressing your comment. Let's wait for travis, and then please go ahead and merge.

PerezHz:
> What bothers me about the error is that it depends on which OS you are running, and I have no idea where to look for such a problem.

Well, at least Linux and Mac agree on the error! Perhaps we should try to trigger a build in a couple of days to see if the problem is related to Julia nightlies?

> I just pushed a new commit addressing your comment. Let's wait for travis, and then please go ahead and merge.

We got the green lights, so I'll just go ahead and merge

PerezHz:
Merged!

PR #78, opened 2019-09-03
add support for saveat expansion

agerlach:
Currently the DifferntialEquations.jl interface to TaylorIntegration.jl supports the taylorinteg(f, x0, t0, tmax, order, abstol, params[=nothing]; kwargs... ) function. I would like to be able to specify a trange using the saveat kwarg. According to [link](http://docs.juliadiffeq.org/stable/basics/common_solver_opts.html#Output-Control-1) if `saveat` is a number it should expand to `tspan[1]:saveat:tspan[2]`

There is also a warning regarding `saveat` support for a `TaylorMethod` that would need update, but  its not clear to me where that resides.


PerezHz:
Hi, @agerlach, thank you for contributing to TaylorIntegration.jl! Since tests are passing I'm in favor of merging this, although coverage has diminished a bit... Could you please add some tests in `test/common.jl` which cover the changes you made? Also, could you please post an example which produces the `TaylorMethod` warning you are seeing, so that we may address it appropriately? Otherwise, this PR is looking good!

agerlach:
This creates the warning, but I see that removing `:saveat` from `warnlist` will solve the problem in common.jl. I will make a commit. 

```julia
using OrdinaryDiffEq
using TaylorIntegration

f(u,p,t) = 1.01*u
u0=1/2
tspan = (0.0,1.0)
prob = ODEProblem(f,u0,tspan)

sol = solve(prob,TaylorMethod(25); saveat = 0.1)
```

agerlach:
> Hi, @agerlach, thank you for contributing to TaylorIntegration.jl!

My pleasure. Thanks for the awesome package. I never heard of a jet transport until I came across your docs. I see a lot of potential for it in my work.

PerezHz:
> My pleasure. Thanks for the awesome package. I never heard of a jet transport until I came across your docs. I see a lot of potential for it in my work.

Hey, it's great to hear that TaylorIntegration.jl is helping you out! If there's anything else we can do to help you, please let us know 😄

src/common.jl
```diff
@@ -68,9 +68,14 @@ function DiffEqBase.solve(
         else
             parse_eqs = true
         end
-        t, vectimeseries = taylorinteg(prob.f.f, prob.u0, prob.tspan[1],
-            prob.tspan[2], alg.order, abstol, prob.p, maxsteps=maxiters,
-            parse_eqs=parse_eqs)
+        if haskey(kwargs, :saveat)
```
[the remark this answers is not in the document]
PerezHz: Perhaps the cases range and vector cases could be handled in some way by dispatch?

lbenet:
Aside from addressing the comments of @PerezHz, I'd like to point out that `DiffEqsBase` has a new release which did break some tests. In #80, among others, we updated the version required by DiffEqsBase`, which is >= 6 now. 

Since #80 is merged (and we plan a new minor release) I  am restarting the tests in travis, so we can check if the changes implemented here are compatible with the new version.

src/common.jl
```diff
@@ -1,6 +1,7 @@
 using DiffEqBase
 
 import DiffEqBase: ODEProblem, solve
+import OrdinaryDiffEq: tstop_saveat_disc_handling
```
agerlach: Added dependency to use the imported function. This is where Chris implements his `saveat` logic

src/common.jl
```diff
@@ -24,13 +25,19 @@ TaylorMethod() = error("Maximum order must be specified for the Taylor method")
...
+    verbose=true,
+    saveat=eltype(tupType)[],
+    abstol = 1e-6,
+    save_start =  isempty(saveat) || saveat isa Number || prob.tspan[1] in saveat,
+    save_end   =  isempty(saveat) || saveat isa Number || prob.tspan[2] in saveat,
     timeseries_errors=true, maxiters = 1_000_000,
     callback=nothing, kwargs...) where
-        {uType, tType, isinplace, AlgType <: TaylorAlgorithm}
+        {uType, tupType, isinplace, AlgType <: TaylorAlgorithm}
+
```
agerlach: Updated the constructed to mimic that used by OrdinaryDiffEq

src/common.jl
```diff
@@ -44,39 +51,69 @@ function DiffEqBase.solve(
...
+        end
     else
         if haskey(kwargs, :parse_eqs)
             parse_eqs = kwargs[:parse_eqs]
         else
             parse_eqs = true
         end
-        if haskey(kwargs, :saveat)
-            t = prob.tspan[1]:kwargs[:saveat]:prob.tspan[2]
-            vectimeseries = taylorinteg(prob.f.f, prob.u0, t, alg.order, abstol, prob.p, maxsteps=maxiters,
```
agerlach: Removing previous implementation

test/common.jl
```diff
@@ -9,19 +9,48 @@ f(u,p,t) = u
...
+    prob = ODEProblem(f!, u0, tspan)
     sol = solve(prob, TaylorMethod(50), abstol=1e-20)
 
     @test norm(sol[end] - u0.*exp(1)) < 1e-12
 end
+
+u0 = 1.0
+tspan = (0.0,5.0)
+saveat_inputs = ([], 0:1:tspan[2], 0:1:(tspan[2]+5), 3:1:tspan[2], collect(0:1:tspan[2]))
+@testset "Test saveat behavior with numbers in common interface" begin
```
agerlach: `saveat_input` lists various use cases: empty, range within tspan, range outside of tspan, range starting after the start time and a list

src/common.jl
PerezHz: Nice!

src/common.jl
PerezHz: OK

src/common.jl
PerezHz: OK

test/common.jl
PerezHz: Great! Thanks for implementing the various cases and adding the corresponding test cases!

PerezHz:
> I think this should address the issues you brought up. I imported the relevant function in OrdinaryDiffEq, so that the logic is the same and should update if Chris makes changes. Additional I added some more rigorous testing to compare the time vectors with that produced by OrdinaryDiffEq.

Sorry for the delay @agerlach... This is LGTM! I was just noting that the master branch on your fork diverged from master at v0.5.1. Could you please rebase the master branch on your fork to the current master here? Once that is taken care of, I'd agree to merge this. @lbenet are there any other changes that we may be missing here?

agerlach:
>  Could you please rebase the master branch on your fork to the current master here? 

My Gitfoo is pretty weak, but I think this should be correct. This is actually my first time doing a pull request.

PerezHz:
> My Gitfoo is pretty weak, but I think this should be correct. 

Looks good!

> This is actually my first time doing a pull request.

It's an honor then! 😄

PerezHz:
I'm merging this, thank you @agerlach for your contribution!

Issue #79, opened 2019-09-05
Test tolerance a bit low


KristofferC:
On the upcoming 1.3 this package errors with

```
Complex dependent variable: Test Failed at /root/.julia/packages/TaylorIntegration/R6yrv/test/taylorize.jl:198
  Expression: norm(abs2(xv1p[end]) - abs2(exact_sol(tv1p[end], cc, cx0)), Inf) < 1.0e-15
   Evaluated: 2.1094237467877974e-15 < 1.0e-15
Stacktrace:
 [1] top-level scope at /root/.julia/packages/TaylorIntegration/R6yrv/test/taylorize.jl:198
 [2] top-level scope at /workspace/srcdir/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
 [3] top-level scope at /root/.julia/packages/TaylorIntegration/R6yrv/test/taylorize.jl:153
Test Summary:              | Pass  Fail  Total
Complex dependent variable |   17     1     18
```

It seems that the test tolerance is put a bit too tightly?

lbenet:
Thanks for reporting! The problem seems already present in Julia 1.3-rc1.0. 

The specific tests are related to some way of optimizing the integration for dependent complex variables (using `@taylorize`), which creates a new (internal) function that should do exactly the same for the integration as the "default" one, just faster. This seems to be ok. Yet, the broken test checks the absolute accuracy of the numerical integration vs the exact solution, and the numerical solution is what has changed. 

@KristofferC Has something (a library or some Julia implementation) related with additions or multiplications of complex numbers changed in Julia 1.3-rc1 in comparison to Julia 1.2? Icouldn't find anything like that in the NEWS.md file...

KristofferC:
Not sure. One way to debug it might be to use Debugger.jl and step through the functions in both Julia versions and see where the difference occurs.

lbenet:
Fixed in #80

PR #80, opened 2019-09-10
Fix a test: increase  tolerance,  using it for all integration steps

lbenet:
The fix was motivated by a broken test in Julia 1.3 but not in Julia 1.2 (#79)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,8 @@ os:
 
 julia:
   - 1.0
-  - 1.1
+  - 1.2
+  - 1.3
   - nightly
 
 notifications:
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 TaylorSeries = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
 
 [compat]
-DiffEqBase = "≥ 5.0.0"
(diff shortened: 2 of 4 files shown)


lbenet:
The problem with the tests seems to be related to `DiffEqBase` which introduced in v6.0.0 some breaking changes with respect to v5.0.

PerezHz:
After taking a look at the changes in DiffEqBase.jl, I realized it was mainly related to the handling of the `callback` argument in `solve`. I was able to fix tests locally (although I'm not 100% sure if it's the right approach); I can push that fix here to see if it helps...

PerezHz:
Just pushed an attempt to fix the issues with DiffEqBase to see how travis and appveyor react

lbenet:
Thanks @PerezHz for pushing the changes. I just pushed another commit to include tests in Julia v1.2 and 1.3.

lbenet:
Once travis finishes, I think we should merge and prepare a new tag. Since we are requiring DiffEqsBase >= 6, I guess a minor version would be appropriate. In that case, I would suggest to get rid of the `info` message displayed about the breaking changes that happened in v0.5. 

@PerezHz What do you think about this?

PerezHz:
> What do you think about this?

I fully agree! I can take care of the new release in another PR after this one is merged

lbenet:
Great! Thanks a lot!

lbenet:
Thanks!

PR #83, opened 2019-09-14
Fixes due to changes in TaylorSeries

lbenet:
This PR fixes few small issues, mostly related with jet transport, which are introduced by https://github.com/JuliaDiff/TaylorSeries.jl/pull/226 (see the discussion in https://github.com/JuliaDiff/TaylorSeries.jl/pull/225). In pass, I rearranged tests to reduce verbose.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -19,7 +19,7 @@ Elliptic = "≥ 0.5.0"
 Espresso = "≥ 0.6.0"
 Reexport = "≥ 0.2.0"
 Requires = "≥ 0.5.2"
-TaylorSeries = "≥ 0.10.0"
+TaylorSeries = "≥ 0.10.1"
 julia = "≥ 1.0.0"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -4,11 +4,11 @@ repo = "https://github.com/PerezHz/TaylorIntegration.jl.git"
 version = "0.6.0"
 
 [deps]
+DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 Espresso = "6912e4f1-e036-58b0-9138-08d1e6358ea9"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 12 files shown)


PerezHz:
Thanks for this contribution and for taking care here of the breaking changes in TaylorSeries! I would imagine that there is going to be a new tagged release after JuliaDiff/TaylorSeries.jl#226 is merged? So should we wait until that is solved?

lbenet:
Rebased to current master...

lbenet:
Tests are passing, though not all have finished (strange?). @PerezHz I think this is ready for review.

PerezHz:
> Tests are passing, though not all have finished (strange?).

Lately I've observed the DiffEqs common interface tests to take a bit longer than usual, but haven't yet figured out what's causing this.

> I think this is ready for review.

Ok, I'll take a look right away!

lbenet:
> Lately I've observed the DiffEqs common interface tests to take a bit longer than usual, but haven't yet figured out what's causing this.

Is it maybe related with the new dependency (`OrdinaryDiffEq.jl`)? Why do we actually require it?

mforets:
> Is it maybe related with the new dependency (OrdinaryDiffEq.jl)? Why do we actually require it?

https://github.com/PerezHz/TaylorIntegration.jl/pull/78#issuecomment-530532453

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/9e0e530dd27fac347ee64b1a98d25ff50b04ec97/src/solve.jl#L411

if it's really just to pull one function, i think it would be preferable to copy it to `TaylorIntegration.jl` (but i haven't checked if this comment makes sense).

lbenet:
I agree with @mforets

PerezHz:
I just pushed a couple of commits which implement the changes suggested by @mforets; let's see how travis responds

PerezHz:
I noticed the slow tests were also due to loading OrdinaryDiffEq in `test/common.jl`; I fixed them so that OrdinaryDiffEq is not required

PerezHz:
Other than that, this PR LGTM, and I agree to merge this once we get the green lights from travis! Thanks @lbenet for the contribution and @mforets for the slow tests fix! Should we tag a patch version after this is merged?

PerezHz:
Merging!

PR #84, opened 2019-09-17
`@taylorize`: add experimental support for `Threads.@threads`

PerezHz:
This PR allows experimental `Threads.@threads` support. Now the user may add `Threads.@threads` before `for` loops in the definition of an ODE when applying the `@taylorize` macro, e.g.:

```julia
@taylorize function f!(dq, q, params, t)
    Threads.@threads for i in 1:10
        dq[i] = q[i]
    end
    nothing
end
```
Also:
```julia
using Base.Threads
@taylorize function f!(dq, q, params, t)
    @threads for i in 1:10 # note there's no `Threads` prefix
        dq[i] = q[i]
    end
    nothing
end
```

I have added some elementary tests which currently only check consistent results on a single thread on travis; right now I'm experimenting with environment variables on travis to actually perform tests on multiple threads.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ git:
 env:
   global:
     - DOCUMENTER_DEBUG=true ## Debug documentation in travis
-    - JULIA_NUM_THREADS=4 ## run Julia on four threads
+    - JULIA_NUM_THREADS=2 ## run Julia on two threads
 
 jobs:
   allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,13 @@ notifications:
 git:
   depth: 99999999
 
+## run Julia on two threads
+env:
+  # JULIA_NUM_THREADS=2
+  matrix:
+    - JULIA_NUM_THREADS=1
+    - JULIA_NUM_THREADS=2
+
 ## Debug documentation in travis
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,9 @@ git:
 
 ## run Julia on two threads
 env:
-  JULIA_NUM_THREADS=4
+  matrix:
+    - JULIA_NUM_THREADS=2
+    - JULIA_NUM_THREADS=4
 
 ## Debug documentation in travis
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,4 +36,5 @@ jobs:
 
 after_success:
   # push coverage results to Coveralls and Codecov
-  - julia -e 'using Pkg; import TaylorIntegration; joinpath(dirname(pathof(TaylorIntegration)), ".."); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder()); Codecov.submit(Codecov.process_folder())'
+  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
+  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,7 @@ git:
 
 ## run Julia on two threads
 env:
-  # JULIA_NUM_THREADS=2
-  matrix:
-    - JULIA_NUM_THREADS=1
-    - JULIA_NUM_THREADS=2
+  JULIA_NUM_THREADS=4
 
 ## Debug documentation in travis
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,16 +17,10 @@ notifications:
 git:
   depth: 99999999
 
-## run Julia on two threads
-env:
-  matrix:
-    - JULIA_NUM_THREADS=2
-    - JULIA_NUM_THREADS=4
-
-## Debug documentation in travis
 env:
   global:
-    - DOCUMENTER_DEBUG=true
+    - DOCUMENTER_DEBUG=true ## Debug documentation in travis
+    - JULIA_NUM_THREADS=4 ## run Julia on four threads
 
 jobs:
   allow_failures:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,6 +14,12 @@ matrix:
   allow_failures:
   - julia_version: nightly
 
+## run Julia on two threads
+environment:
+  matrix:
+    - JULIA_NUM_THREADS: 1
+    - JULIA_NUM_THREADS: 2
+
 # branches:
 #   only:
 #     - master
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,11 +2,17 @@ environment:
   matrix:
   - julia_version: 1.0
   - julia_version: 1.3
+    JULIA_NUM_THREADS: 1 # single thread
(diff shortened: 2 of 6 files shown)


lbenet:
Thanks for this PR!

A naive question: Does this work for ODEs which couple the different degrees of freedom?

PerezHz:
> A naive question: Does this work for ODEs which couple the different degrees of freedom?

I've done some locals tests which show that it does; I'll try to add those here to see if they pass on travis

PerezHz:
> A naive question: Does this work for ODEs which couple the different degrees of freedom?

@lbenet this is ready for review. I just added a test case with coupled degrees of freedom and things seem to be working alright. I added a couple of `@time`s in the tests just to see how performance is doing on travis, but I'll remove them in the next commits I make here before merging.

PerezHz:
Tests are passing on Linux and Windows, but somehow tests are failing on macOS (julia 1.2 and 1.3), I will take a look into this

PerezHz:
Tests on julia nightly (macOS) are failing as well...

PerezHz:
I was able to fix the error (it was related to an auto-assignment); tests are passing again 👍

PerezHz:
Rebased to current master.

lbenet:
There seems a conflict with appveyor.yml; probably due to #102. Can you rebase to current master?

.travis.yml
```diff
@@ -36,4 +36,5 @@ jobs:
 
 after_success:
   # push coverage results to Coveralls and Codecov
-  - julia -e 'using Pkg; import TaylorIntegration; joinpath(dirname(pathof(TaylorIntegration)), ".."); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder()); Codecov.submit(Codecov.process_folder())'
+  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
+  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
```
lbenet: Don't you need `import TaylorIntegration; joinpath(dirname(pathof(TaylorIntegration)), "..");` in each line, after `using Pkg`?

appveyor.yml
```diff
@@ -2,7 +2,9 @@ environment:
   matrix:
   - julia_version: 1.0
   - julia_version: 1.2
+  - julia_version: 1.3 # single thread
   - julia_version: 1.3
+    JULIA_NUM_THREADS: 2 # two threads
   - julia_version: nightly
```
lbenet: (This file needs to be rebase to current master.) 

I think it is advisable to test this functionality in Julia 1.3 and beyond.

docs/src/taylorize.md
```diff
@@ -208,6 +208,8 @@ list some limitations and advices.
 - Use `local` for internal parameters (simple constant values); this improves
   performance. Do not use it if the variable is Taylor expanded.
 
+- `@taylorize` supports multi-threading via `Threads.@threads`.
```
lbenet: I think we should signal that this is experimental. Do we know if using `@threads` would be thread-safe?

test/taylorize.jl
```diff
@@ -5,6 +5,7 @@ using Test
 using LinearAlgebra: norm
 using InteractiveUtils: methodswith
 using Elliptic
+using Base.Threads
```
lbenet: We may need to add Base.Threads to the `[deps]` section, as well as to `[extras]` and `[targets]` in `Project.toml`.

.travis.yml
```diff
@@ -17,10 +17,10 @@ notifications:
 git:
   depth: 99999999
 
-## Debug documentation in travis
 env:
   global:
-    - DOCUMENTER_DEBUG=true
+    - DOCUMENTER_DEBUG=true ## Debug documentation in travis
+    - JULIA_NUM_THREADS=2 ## run Julia on two threads
```
lbenet: Do you know if there is a *programmatic* way of changing the number of threads? I am thinking in a way to avoid setting an ENV variable...

PerezHz: I'm not aware of a programmatic way to set it programatically... Currently Julia reads the number of threads from the env variable `JULIA_NUM_THREADS`, which must be set *before* Julia starts. It can be set from within Julia, but then the Julia session should be restarted so that the changes take effect.

docs/src/taylorize.md
PerezHz: > I think we should signal that this is experimental. 

Ok, I'll add that warning.

>  Do we know if using @threads would be thread-safe?

That is up to the user, since that depends on each particular ODE, so we cannot guarantee thread-safety in advance. Perhaps I should document this as well.

appveyor.yml
PerezHz: Thanks, just rebased to current master

PerezHz: ... and added corresponding env variable settings for julia1.4 and nightly

docs/src/taylorize.md
lbenet: Yes, it's important to explicit mention it.

.travis.yml
lbenet: Thanks for the clarification!

.travis.yml
PerezHz: Instructions in [Coverage.jl](https://github.com/JuliaCI/Coverage.jl#tracking-coverage-with-codecovio) say that it's not necessary. Actually, I was just reading that it may be enough to add `codecov: true` and `coveralls: true` to .travis.yml instead of these two lines.

test/taylorize.jl
PerezHz: Thanks, I added Threads in the corresponding sections of Project.toml

.travis.yml
lbenet: Fair enough; go ahead and make the changes to see :-)

docs/src/taylorize.md
PerezHz: Ok, both warnings are now documented, thanks!

PerezHz:
Many thanks for taking a look into this, @lbenet!

> I've checked the changes proposed in this PR. It seems that everything is consistent, except for some minor observations. In any case, this PR needs to be rebased to master.

Just rebased to master and pushed a couple of commits which address the main points of your review; please let me know if there is anything that could be improved further.

lbenet:
There seems to be a problem with the tests, perhaps related to my proposal of including Threads in Project.toml. @PerezHz, can you remove them and see if things work properly then?

PerezHz:
> There seems to be a problem with the tests, perhaps related to my proposal of including Threads in Project.toml. @PerezHz, can you remove them and see if things work properly then?

Sorry for the late reply, and thanks for checking this! Just pushed a commit which reverts the changes in Project.toml; let's see how travis and appveyor react

PerezHz:
Tests seem to be passing, except for a codecov error which is not shown by coveralls... should we merge this and possibly tag a new patch/minor version? What do you think @lbenet?

lbenet:
Thanks a lot @PerezHz . I'll merge it now.

lbenet:
@PerezHz What about tagging a new (patch) version?

PerezHz:
Thanks for merging and tagging a new patch version!

PR #88, opened 2019-10-31
Upper bound versions of dependencies

lbenet:
Related to [this comment](https://github.com/JuliaRegistries/General/pull/4612#issuecomment-545130953).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,13 +14,13 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 TaylorSeries = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
 
 [compat]
-DiffEqBase = "≥ 6.0.0"
-Elliptic = "≥ 0.5.0"
-Espresso = "≥ 0.6.0"
-Reexport = "≥ 0.2.0"
-Requires = "≥ 0.5.2"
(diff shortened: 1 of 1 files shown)


PerezHz:
Thanks for this PR! LGTM; should we go ahead and merge? Since the only changed file is Project.toml, there shouldn't be any changes in coverage, so perhaps we should just dismiss the coveralls report here.

lbenet:
Sorry @PerezHz for the long delay, and thanks for taking a look into this. 

I agree, we should not take in this case too seriously the failure in the coverage. Yet, maybe we should wait for #89 before merging this, and  tag a new patch version (trying to use the automatic merging machinery).

PerezHz:
Rebased to current master

lbenet:
#89 is merged. Do you think we can go ahead with this one? And perhaps also tag a new --perhaps patch, but it may not work the automatic merging-- version?

PerezHz:
Just merged this PR, maybe I can try now to tag a new patch version?

lbenet:
Thanks!

I do not think it will work as an automatic merge, but I don't think it should be a minor version

PerezHz:
Just opened #90 in order to tag v0.7.1

lbenet:
Thanks again!

PR #91, opened 2020-02-06
Change convention for the event function `g`

lbenet:
I am opening this to propose a different return convention for the function `g`, which is used in some methods which involve root-finding/event-detection functions. The motivation is to avoid the possibility of an error (which I describe below), which is solved by defining a different convention on what should the function `g` return.

First, I describe the possibility of an error: Following the example on [Poincare maps in the docs](https://perezhz.github.io/TaylorIntegration.jl/latest/root_finding/#Monte-Carlo-simulation-1), the function `g` returns either `x[1]` if `x[3]>0`, or `nothing`. (The Poincaré map is defined as `x[1]=0` with `x[3]>0`.) Then, [in this line](https://github.com/PerezHz/TaylorIntegration.jl/blob/8e031c4e77085f5567b7a504582f315329c869a7/src/rootfinding.jl#L195) and also [here](https://github.com/PerezHz/TaylorIntegration.jl/blob/8e031c4e77085f5567b7a504582f315329c869a7/src/rootfinding.jl#L296)), we may end up evaluating `zero(nothing)`, which throws a `MethodError`. 

(In pass, I'd like to note that both functions have slightly different arguments for the function `g` which I think should read in both cases `g(dx, x, params, t)`.)

My proposal is that `g` should return a tuple similar to `Tuple{Bool, Taylor1{T}}`, the first entry being the Boolean `x[3]>0` (for the example in the docs,) and the second, `x[1]`.

### DIFF
--- a/docs/src/root_finding.md
+++ b/docs/src/root_finding.md
@@ -3,7 +3,7 @@
 In this example, we shall illustrate how to construct a Poincaré map associated
 with the surface of section ``x=0``, ``\dot x>0``, for ``E=0.1025`` for the
 [Hénon-Heiles system](https://en.wikipedia.org/wiki/Hénon–Heiles_system). This is
-equivalent to find the roots of an appropriate function `g(t, x, dx)`. We
+equivalent to find the roots of an appropriate function `g(dx, x, params, t)`. We
 illustrate the implementation using many initial conditions (Monte
 Carlo like implementation), and then compare the results
 with the use of [jet transport techniques](@ref jettransport).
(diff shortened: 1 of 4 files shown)


lbenet:
@PerezHz Let me know what you think about this.

lbenet:

PerezHz:
@lbenet this is great, thanks! I just reviewed the changes and they LGTM... CI is giving green lights, so is this ready to be merged?

lbenet:
I guess there are only two questions 
- Is if the convention I propose is ok, or should we perhaps opt for returning a `Tuple{Taylo1{T}, Bool}`? Either choice is good for me.
- Should we tag a new (minor) version?

Aside from that, it is ready to e merged.

PerezHz:
> Is if the convention I propose is ok, or should we perhaps opt for returning a Tuple{Taylor1{T}, Bool}? Either choice is good for me.

Yeah I agree, either choice is as good as the other one.

> Should we tag a new (minor) version?

Yes I think we should; I'll handle that after merging this!

PerezHz:
Merged!

lbenet:
Thanks a lot!

PR #97, opened 2020-02-13
Implement `elseif` blocks with macro

### LINKED ISSUE #96: taylorize macro does not recognize `elseif` statements
In current master, when trying to use `@taylorize` with the following function definition

```julia
@taylorize function f!(dq, q, params, t)
    for i in 1:10
        if i == 1
            dq[i] = 2q[i]
        elseif i == 2
            dq[i] = -q[i]
        else
            dq[i] = q[i]
        end
    end
    nothing
end
```

An error occurs:

```
ERROR: LoadError: ArgumentError: collection must be non-empty
```

If, instead, the `elseif` statement is removed, things work fine:
```julia
@taylorize function f!(dq, q, params, t)
    for i in 1:10
        if i == 1
            dq[i] = 2q[i]
        else
            dq[i] = q[i]
        end
    end
    nothing
end
```

lbenet:
Fixes #96

### DIFF
--- a/src/parse_eqs.jl
+++ b/src/parse_eqs.jl
@@ -262,7 +262,7 @@ function _preamble_body(fnbody, fnargs, debug=false)
     debug && (println("------ _parse_newfnbody! ------");
         @show(newfnbody); println(); @show(preamble); println();
         @show(v_vars); println(); @show(d_indx); println();
-        @show(v_newindx); println())
+        @show(v_assign); println(); @show(v_newindx); println())
 
     # Include the assignement of indexed auxiliary variables
     defspreamble = _defs_preamble!(preamble, fnargs,
@@ -792,8 +792,8 @@ function _recursionloop(fnargs, retvar)
     ll = length(fnargs)
     if ll == 3
 
-        rec_preamb = sanitize( :( $(fnargs[2])[1] = $(retvar)[0] ) )
-        rec_fnbody = sanitize( :( $(fnargs[2])[ordnext] = $(retvar)[ord]/ordnext ) )
+        rec_preamb = sanitize( :( $(fnargs[1])[1] = $(retvar)[0] ) )
+        rec_fnbody = sanitize( :( $(fnargs[1])[ordnext] = $(retvar)[ord]/ordnext ) )
         #
     elseif ll == 4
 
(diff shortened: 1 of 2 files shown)


lbenet:
I've included `continue` and `break` statements as well.

The solution does not shine because its elegance, but I think it works.

lbenet:
I still need to provide some tests for the new cases...

lbenet:
@PerezHz I have added some tests, in what I believe is the *expected* way the macro acts. Can you please take a look on those, especially in the parts in which not-recognized functions are just copied?

src/parse_eqs.jl
```diff
@@ -340,9 +340,17 @@ function _newfnbody(fnbody, fnargs, d_indx)
...
                 append!(v_arraydecl, tmp_arraydecl)
             elseif ex_head == :if
+                # The first argument of an `if` expression is the condition, the
+                # second one is the block the condition is true, and the third
+                # is the `else`-block or an `ifelse`-block.
                 push!(newfnbody.args, Expr(:if, ex.args[1]))
                 for exx in ex.args[2:end]
-                    ifbody, tmp_newindx, tmp_arraydecl = _newfnbody( exx, fnargs, d_indx)
+                    if exx.head == :elseif
+                        exxx = Expr(:block, Expr(:if, exx.args[1].args[2], exx.args[2:end]...))
```
PerezHz: Nice workaround! I still can't figure out why `:elseif` blocks can't be treated straight-forward as `:if` blocks 🤷‍♂

test/taylorize.jl
```diff
@@ -129,7 +143,7 @@ using Elliptic
 
     # Pendulum integrtf = 100.0
     @testset "Integration of the pendulum and DiffEqs interface" begin
-        @taylorize function pendulum!(dx, x, p, t)
+        @taylorize function pendulum!(dx::Array{T,1}, x::Array{T,1}, p, t) where {T}
```
PerezHz: Are we just testing type annotations here in the function definition, is there another reason to add them?

lbenet: Yes, you are right. Internally, we simply ignore the type annotations, but there was no test.

lbenet:
Thanks @PerezHz for the review. As far as I am concerned, I think this is ready for merging, specially if it works fine for you. So, please, go ahead and merge this.

PerezHz:
Merging! Again, thanks a lot @lbenet for taking care of this!

PR #98, opened 2020-02-20
Add tests to check that the parsed jetcoeffs! method is used

lbenet:
... including methods in which it fails, despite the macro works.

### DIFF
--- a/src/explicitode.jl
+++ b/src/explicitode.jl
@@ -242,6 +242,47 @@ function taylorstep!(f!, t::Taylor1{T}, x::Vector{Taylor1{U}},
     return δt
 end
 
+
+
+"""
+    _determine_parsing!(parse_eqs::Bool, f, t, x, params)
+    _determine_parsing!(parse_eqs::Bool, f, t, x, dx, params)
+
+Check if the parsed method of `jetcoeffs!` exists and check it
+runs without error.
+"""
+function _determine_parsing!(parse_eqs::Bool, f, t, x, params)
+    parse_eqs = parse_eqs &&
+        !isempty(methodswith(Val{f}, TaylorIntegration.jetcoeffs!))
+    if parse_eqs
+        try
+            jetcoeffs!(Val(f), t, x, params)
+        catch
+            @warn("""Unable to use the parsed method of `jetcoeffs!`
+            despite of having `parse_eqs=true`, due to some internal error.
+            Using `parse_eqs = $false`""")
+            parse_eqs = false
+        end
+    end
+    return parse_eqs
+end
+function _determine_parsing!(parse_eqs::Bool, f, t, x, dx, params)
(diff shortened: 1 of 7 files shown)


lbenet:
This was motivated because I noticed that, sometimes, despite of `parse_eqs=true` and that the macro seems to work fine, the actual integration took place with `parse_eqs=false`.

lbenet:
@PerezHz Should we warn/inform the user which method is actually used for the integration?

PerezHz:
> @PerezHz Should we warn/inform the user which method is actually used for the integration?

Yes, I agree, we should inform the user in case the value they select for `parse_eqs` is changed internally by `taylorinteg`. E.g., if a user calls `taylorinteg` with `parse_eqs=true` and `taylorinteg` for some reason changes `parse_eqs=false`, then we should throw a warning

lbenet:
It took me some time to realize that I *also* needed to add `InteractiveUtils` in the `[extras]` section of `Project.toml`. In doing this, I noticed that the tests are broken for Julia 1.1, in what I think is due to a scoping issue, that I didn't try to solve. I include this info here as a reference.

src/explicitode.jl
```diff
@@ -242,6 +242,47 @@ function taylorstep!(f!, t::Taylor1{T}, x::Vector{Taylor1{U}},
...
+            despite of having `parse_eqs=true`, due to some internal error.
+            Using `parse_eqs = $false`""")
+            parse_eqs = false
+        end
+    end
+    return parse_eqs
+end
+function _determine_parsing!(parse_eqs::Bool, f, t, x, dx, params)
+    parse_eqs = parse_eqs &&
+        !isempty(methodswith(Val{f}, TaylorIntegration.jetcoeffs!))
```
PerezHz: Nice! Using `methodswith` this way, we only look for `jetcoeffs!` methods related to `f`, right?

src/explicitode.jl
```diff
@@ -242,6 +242,47 @@ function taylorstep!(f!, t::Taylor1{T}, x::Vector{Taylor1{U}},
...
+runs without error.
+"""
+function _determine_parsing!(parse_eqs::Bool, f, t, x, params)
+    parse_eqs = parse_eqs &&
+        !isempty(methodswith(Val{f}, TaylorIntegration.jetcoeffs!))
+    if parse_eqs
+        try
+            jetcoeffs!(Val(f), t, x, params)
+        catch
+            @warn("""Unable to use the parsed method of `jetcoeffs!`
```
PerezHz: Thanks for adding this warning!

src/explicitode.jl
lbenet: Yes! I think this is much better than before

lbenet:
Thanks for the review! I'll go ahead and merge this

PR #102, opened 2020-03-04
Fix displayed equations in the docs

lbenet:


lbenet:
I've also added the possibility of inspecting the docs produced by pull requests.

lbenet:
The docs produced by this PR can be found [here](https://perezhz.github.io/TaylorIntegration.jl/previews/PR102). 

I think everything works as expected, but I find extremely slow the [Kepler example](https://perezhz.github.io/TaylorIntegration.jl/previews/PR102/kepler/); this also happens in the current stable and dev docs. I suspect that this is due to the enormous number of points plotted for the orbit; I'll push a new commit using only 10000 points in that plot. Another possibility is to revert the latex changes and use MathJax as HTML renderer. Any preferences @PerezHz ?

lbenet:
I reduced the number of points plotted in the figures for the Kepler example. This helps for rendering the page, which is now again usable. 

I think this is ready @PerezHz.

lbenet:
Merging.

Issue #106, opened 2020-06-22
Warning: incremental compilation (maybe) broken when using `@taylorize` in package


PerezHz:
When using `@taylorize` inside a module, a warning appears about incremental compilation "maybe" being "fatally broken". I wasn't able to reproduce the problem when defining a module in the REPL, but figured out the following MWE:

I created the [`MyPkg`](https://github.com/PerezHz/MyPkg) julia package, which uses `@taylorize` to define an ODE. Then, `MyPkg` may be `git clone`'d on an empty folder, and from `MyPkg` root dir start a julia REPL and do:
```julia
# first, in an empty folder do:
# git clone https://github.com/PerezHz/MyPkg
# cd MyPkg
# julia ### open a Julia REPL
# then, in the REPL do:
import Pkg
Pkg.activate(".") # activate environment on MyPkg root dir
using MyPkg # warning appears
```
The warning messages that appear are:
```
[ Info: Precompiling MyPkg [87ba5f82-8271-4dfc-8d8f-dcd99beeaf9d]
WARNING: eval into closed module TaylorIntegration:
:-
  ** incremental compilation may be fatally broken for this module **

WARNING: eval into closed module TaylorIntegration:
:^
  ** incremental compilation may be fatally broken for this module **

WARNING: eval into closed module TaylorIntegration:
typeof(MyPkg.xdot2)()
  ** incremental compilation may be fatally broken for this module **

WARNING: eval into closed module TaylorIntegration:
nothing
  ** incremental compilation may be fatally broken for this module **
```

So far, I've reproduced the error on julia 1.3 and 1.4 locally and on a remote machine.

The incremental compilation warnings appear the first time `MyPkg` is loaded via `using MyPkg`, and do not appear on subsequent loadings, until `MyPkg` is changed internally (e.g., by uncommenting the `MyPkg.greet()` function), and `using MyPkg` is ran again so that the code is re-compiled. Also, as far as I can tell, things seem to work fine (`taylorinteg` works fine, etc.) even when the warnings appear, but I don't know if it's safe to live with these warnings hanging around.

A workaround is to set `__precompile__(false)`, but then `MyPkg` is compiled each time it is loaded, and compilation times may become exceptionally slow (~ 10 min) in some cases where the defined ODE involves many operations.

[...]

PR #112, opened 2020-10-07
Update taylorize.md

mforets:
see https://github.com/PerezHz/TaylorIntegration.jl/issues/106

### DIFF
--- a/docs/src/taylorize.md
+++ b/docs/src/taylorize.md
@@ -219,3 +219,41 @@ It is recommended to skim `test/taylorize.jl`, which implements different
 cases.
 
 Please report any problems you may encounter.
+
+## Usage inside a module
+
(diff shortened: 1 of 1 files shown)


lbenet:
@mforets First of all, thanks for addressing documenting #106. In view of #113 the  trick of copy-pasting the definition of the macro is not useful anymore. Yet, I think it is worth keeping some reference about the problem in the docs. Would you care about fixing this and we merge it (together with #113).

mforets:
ok, i wrote this docs PR because i also hit this issue in another package and this solved it!

> Would you care about fixing this and we merge it (together with #113).

 feel free to close this PR :)  not sure what needs to be documented on the TaylorIntegration.jl side.

lbenet:
Thanks anyway for the effort! 

As mentioned, #113 should fix the problem. I just merged, so if you could test the current version of master that would be great!

I'll close this!

PR #113, opened 2020-10-08
Fix #106

### LINKED ISSUE #106: Warning: incremental compilation (maybe) broken when using `@taylorize` in package

PerezHz:
Since #106 was last closed, I've been able to trace that issue to [this line of code](https://github.com/dfdx/Espresso.jl/blob/76c899cbd19df6f18e10875439591368fbc9b679/src/utils.jl#L151) in Espresso.jl, where the use of `Core.eval` leads to a silent error when using `@taylorize` from inside a module via the use of `ExGraph` [here](https://github.com/PerezHz/TaylorIntegration.jl/blob/1f36a47362bfe4d74022f257d20f5a78b1eea110/src/parse_eqs.jl#L412). For the time being, I have been able to workaround this issue by re-implementing the relevant part of Espresso in TaylorIntegration without using `eval`, but a better solution might be to report this directly there to look for a better solution. So, current changes are not elegant, but they get the job done, as far as I could tell.

### DIFF
--- a/test/TestPkg/src/TestPkg.jl
+++ b/test/TestPkg/src/TestPkg.jl
@@ -1,0 +1,183 @@
+module TestPkg
+
+# __precompile__(false)
+
+using TaylorIntegration
+# using BenchmarkTools
+using InteractiveUtils
+
+order = 25
+N = 200
+x0 = 10randn(2N)
+μ = 1e-7rand(N)
+x0[N] = - sum(x0[1:N-1].*μ[1:N-1])/μ[N]
+x0[2N] = - sum(x0[N+1:2N-1].*μ[1:N-1])/μ[N]
+
+@taylorize xdot1(x, p, t) = b1-x^2
+
+@taylorize function f1!(dq, q, params, t)
+    local N = Int(length(q)/2)
+    local _eltype_q_ = eltype(q)
+    local μ = params
+    X = Array{_eltype_q_}(undef, N, N)
+    accX = Array{_eltype_q_}(undef, N) #acceleration
+    for j in 1:N
+        accX[j] = zero(q[1])
+        dq[j] = q[N+j]
+    end
+    #compute accelerations
+    for j in 1:N
+        for i in 1:N
+            if i == j
+            else
+                X[i,j] = q[i]-q[j]
+                temp_001 = accX[j] + (μ[i]*X[i,j])
+                accX[j] = temp_001
+            end #if i != j
+        end #for, i
+    end #for, j
+    for i in 1:N
+        dq[N+i] = accX[i]
+    end
+    nothing
+end
+
+function f2!(dq, q, params, t)
+    local N = Int(length(q)/2)
+    local _eltype_q_ = eltype(q)
+    local μ = params
+    X = Array{_eltype_q_}(undef, N, N)
+    accX = Array{_eltype_q_}(undef, N) #acceleration
+    for j in 1:N
+        accX[j] = zero(q[1])
+        dq[j] = q[N+j]
+    end
+    #compute accelerations
+    for j in 1:N
+        for i in 1:N
+            if i == j
+            else
+                X[i,j] = q[i]-q[j]
+                temp_001 = accX[j] + (μ[i]*X[i,j])
+                accX[j] = temp_001
+            end #if i != j
+        end #for, i
+    end #for, j
+    for i in 1:N
+        dq[N+i] = accX[i]
+    end
+    nothing
+end
+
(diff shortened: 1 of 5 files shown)


lbenet:
Thanks for finding the error? Why there is no manifestation of the error? Do the fix proposed in https://github.com/dfdx/Espresso.jl/issues/49 solves  the problems?

PerezHz:
> Thanks for finding the error? Why there is no manifestation of the error? Do the fix proposed in [dfdx/Espresso.jl#49](https://github.com/dfdx/Espresso.jl/issues/49) solves the problems?

The error happens silently since the call to `ExGraph` is inside a `try ... catch ... end` block; after testing the solution proposed in [dfdx/Espresso.jl#49](https://github.com/dfdx/Espresso.jl/issues/49) it seems to be working without changing the code in `src/parse_eqs.jl`, so it fortunately looks like most of those changes won't be needed once a new Espresso.jl version is released. Still, I think it's important to have this tested, which is what I'm trying to do add now.

PerezHz:
@lbenet, regarding the silent failure, I'm proposing to ditch the `try...catch...end` block which called `ExGraph` in `parse_eqs.jl`: I found it is not needed anymore (i.e., it does not falling back to the `catch` anymore) and instead I think it would be useful to know if a call to `ExGraph` fails for any reason, what do you think? I also added some tests regarding the use of `@taylorize` within Julia packages. Tests are passing with Espresso.jl v0.6.1 (thanks @dfdx!), so after a trip to a famous mythological labyrinth in Minos, this is ready for review...

test/TestPkg/src/TestPkg.jl
```diff
@@ -0,0 +1,183 @@
+module TestPkg
```
lbenet, line 1: I fully agree to include the tests you have implemented here. I suggest to include few comments (perhaps outside from `TestPkg` simply to describe what you are testing. Our memory is finite and we are prone to forget!

lbenet (approved):
First of all, thanks for finding out the true problem that was causing #106!

This PR LGTM. I have only added one comment, to add some comments about what `MyPkg` has inside, that we want to test; few lines simply counteracting of finite memory.

lbenet (approved):
Thanks a lot!

lbenet:
Merging!

Issue #109, opened 2020-07-23
Compatibility with DynamicalODEProblem


SebastianM-C:
The current implementation is incompatible with DynamicalODEProblem specifications from DiffEq since
https://github.com/PerezHz/TaylorIntegration.jl/blob/9907e81f696d98ae5225596d3514eb15bd6889d0/src/common.jl#L51

is specific to `ODEFunction`s and will fail for `DynamicalODEFunction`s (which have `f1` and `f2`). I tried to address this in #108

PerezHz:
@SebastianM-C in #114 we are proposing a new implementation for the common interface with DiffEq. Although surely much can be improved, so far it looks promising; perhaps the new approach can help with this issue? Feel free to take a look and let us know what you think 😄

SebastianM-C:
Thanks for keeping me in the loop. From a first look I think it would remove much of the previous special casing in dispatch and thus helps solve most of the problems. I'll take a closer look the following days.

lbenet:
@SebastianM-C #114 has been merged. Is this still an issue?

SebastianM-C:
No, I don't think so.

Issue #103, opened 2020-04-17
Distinguish `tspan` and `saveat` in taylorintegration


lbenet:
Currently, we have different ways to set the initial and final times for the integration, which defines which data is returned. DifferentialEquations distinguishes `tspan` (a tuple for the initial and final time) from `saveat`, which turns to be pretty handy.

PR #114, opened 2020-10-14
Re-implement common interface bindings

PerezHz:
This PR proposes to re-implement the common interface bindings to the `DifferentialEquations.jl` ecosystem essentially by adding `TaylorMethod` as an ODE solver method of `OrdinaryDiffEq.jl` in the sense `TaylorAlgorithm <: OrdinaryDiffEqAdaptiveAlgorithm`. So the proposed approach is to extend `initialize!` and `perform_step!` methods in `OrdinaryDiffEq.jl`, while handling step size control via corresponding method extensions of `stepsize_controller!` and `step_accept_controller!`. By doing this, calling `solve` yields the same results as a direct call to `taylorinteg` with comparable performance, while allowing the use of a larger set of keywords with `TaylorMethod`, including `saveat`, `tstops` and even `callback` (currently, discrete callbacks seem to work okay while continuous callbacks do not).

On the `TaylorIntegration.jl` side, I extended `jetcoeffs!` and `stepsize`, as well as `TaylorSeries.evaluate!` to allow use of high-dimension arrays. Following this strategy, current tests in `master` pass, and also added some more tests. In particular, this PR might help to find a way out of #103 and #109.

There are some issues remaining: besides the continuous callbacks (which I'm currently trying to sort out) I'm not completely sure how to handle the FSAL stuff, but since it is used in 3rd-order Hermite interpolation in `OrdinaryDiffEq`, I added FSAL-related updates in `initialize!` and `perform_step!`, although I'm not sure if I did it right. Also, I left most of the keywords in the warning list `warnkeywords`, although a better approach might now be the inverse: to ditch the keyword warning list altogether (or at least most of it?), and handle any issues that will arise in the future.

Surely there are lots of things that can be improved, so feedback is appreciated!

cc: @ChrisRackauckas @lbenet


PerezHz:
Updates: I found a way, thanks to @SebastianM-C's work in #108, to finally handle `DynamicalODEProblem`s with TaylorIntegration.jl. A thing to note is that use of `@taylorize` for speedup is currently not allowed for DynamicalODEProblem. Anyway, with the latest changes this PR fixes #109. We can always take care of performance for DynamicalODEProblem, via use of `@taylorize`, in a future PR.

lbenet:
Thanks for the great effort! 

Please ping me, and perhaps @SebastianM-C too, whenever you think this is ready for review.

PerezHz:
> Please ping me, and perhaps @SebastianM-C too, whenever you think this is ready for review.

The latest update is that I think I found a way to make continuous and vector callbacks work (discrete callbacks were already working), which I hope will help users of TaylorIntegration.jl through the common interface.

I'm still not 100% percent sure about the FSAL thing, but things appear to be working well, so... 🤷‍♂️. The keyword warning list perhaps needs to be updated in some way as well, but here's where some feedback might help to know what's the best way to proceed going forward. I think we should add support, respectively, for in-place format in scalar problems and out-of-place format in array problems, but we might tackle that in another PR. Finally, in the tests some funny warnings started to appear:
```
WARNING: Method definition vec(Number) in module FiniteDiff at
C:\Users\appveyor\.julia\packages\FiniteDiff\EBys0\src\jacobians.jl:128 
overwritten in module DiffEqDiffTools at
C:\Users\appveyor\.julia\packages\DiffEqDiffTools\uD0fb\src\jacobians.jl:114.
```
but that seems to be related to FiniteDiff and DiffEqDiffTools, since we're not overwriting any `vec` methods here.

Otherwise, this PR is ready for review!

cc: @lbenet @SebastianM-C

src/common.jl
```diff
@@ -1,194 +1,283 @@
...
-                dd = reshape(du, sizeu); uu = reshape(u, sizeu);
-                f(dd, uu, p, t); u = vec(uu); du = vec(dd); 0)
-        u0 = vec(prob.u0)
-        if isempty(saveat_vec)
-            t, vectimeseries = taylorinteg(f!, u0, prob.tspan[1], prob.tspan[2],
-                alg.order, abstol, prob.p, maxsteps=maxiters, parse_eqs=false)
+    sizeu = size(prob.u0)
+    # This if block handles DynamicalODEProblems
+    # credit to @SebastianM-C for coming up with this solution
+    if prob.f isa DynamicalODEFunction
```
SebastianM-C, line 192: From what I understand, this special casing is needed because the `jetcoeffs!` function calls the user function with `xaux` which is an `Array`. I was wondering if it would be possible to just use the `DiffEqFunction` interface (i.e. call `prob.f` with `f(du,u,p,t)` and `u` with the same type as `prob.u0`), but I'm not sure I understand how the `jetcoeffs!` code works and if the above suggestion would make sense.

SebastianM-C (commented):
Thanks for adding support for `DynamicalODEProblems`. I was thinking if a more general approach where the conversion to `ODEProblem` is no longer required would be possible.

src/common.jl
PerezHz: Thank you for your review and your suggestions! You're right, it'd be nice if we didn't have to convert everything to an `ODEProblem`... after reading your comment here I gave it a shot, and I think just got to overcome this conversion, at least for the in-place case (I should push that commit in a couple of minutes). The out-of-place version won't work as of now, since `TaylorIntegration.jl` currently does not support oop for `prob.u0 <: AbstractArray`. But once we add that, perhaps in another PR, we'll be able to get rid of most of the `if`s and `else`s, while avoiding the conversion from `DynamicalODEProblem` to `ODEProblem`.

PerezHz:
Just pushed a commit which fixed an error in the casing for out-of-place DynamicalODEProblem (the returned value from f wasn't being passed correctly) and added some tests.

PerezHz:
@SebastianM-C I was thinking... is there a way to convert an out-of-place `DynamicalODEProblem` to an in-place `DynamicalODEProblem`? Since we now can handle in-place `DynamicalODEProblem` if somehow we were able to do that conversion, that way we could avoid the conversion to `ODEProblem` for out-of-place `DynamicalODEProblem`...

PerezHz:
I was trying stuff like
```julia
if prob.f isa DynamicalODEFunction && !isinplace
    f1! = (dv, v, u, p, t) -> (dv .= prob.f.f1(v, u, p, t); 0)
    f2! = (du, v, u, p, t) -> (du .= prob.f.f2(v, u, p, t); 0)
    prob.f = DynamicalODEFunction{true}(f1!, f2!)
    # ...
```
which is similar to what we do for out-of-place `ODEProblem`, but then I get an error:
```
ERROR: MethodError: Cannot `convert` an object of type
  DynamicalODEFunction{true,ODEFunction{true,typeof(iip_ṗ),LinearAlgebra.UniformScaling{Bool{}},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{}},ODEFunction{true,typeof(iip_q̇),LinearAlgebra.UniformScaling{Bool{}},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{}},LinearAlgebra.UniformScaling{Bool{}},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{}} to an object of type
  DynamicalODEFunction{false,ODEFunction{false,typeof(oop_ṗ),LinearAlgebra.UniformScaling{Bool{}},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{}},ODEFunction{false,typeof(oop_q̇),LinearAlgebra.UniformScaling{Bool{}},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{}},LinearAlgebra.UniformScaling{Bool{}},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{},Nothing{}}
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:171
  DynamicalODEFunction{false,ODEFunction{false,typeof(oop_ṗ),LinearAlgebra.UniformScaling{Bool},Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing},ODEFunction{false,typeof(oop_q̇),LinearAlgebra.UniformScaling{Bool},Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing},LinearAlgebra.UniformScaling{Bool},Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing}(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any) where {iip, F1, F2, TMM, Ta, Tt, TJ, JVP, VJP, JP, SP, TW, TWt, TPJ, S, TCV} at /Users/Jorge/.julia/packages/DiffEqBase/3iigH/src/diffeqfunction.jl:64
Stacktrace:
 [1] setproperty!(::ODEProblem{RecursiveArrayTools.ArrayPartition{Float64,Tuple{SArray{Tuple{2},Float64,1,2},SArray{Tuple{2},Float64,1,2}}},Tuple{Float64,Float64},false,DiffEqBase.NullParameters,DynamicalODEFunction{false,ODEFunction{false,typeof(oop_ṗ),LinearAlgebra.UniformScaling{Bool},Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing},ODEFunction{false,typeof(oop_q̇),LinearAlgebra.UniformScaling{Bool},Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing},LinearAlgebra.UniformScaling{Bool},Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing},Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}},DiffEqBase.StandardODEProblem}, ::Symbol, ::Function) at ./Base.jl:34
 [2] top-level scope at REPL[35]:1
```

In the tests, if I try

```julia
oop_prob.f = iip_prob.f
```
the same error shows up.

test/common.jl
```diff
@@ -235,11 +241,11 @@ using StaticArrays
         function oop_ṗ(p, q, params, t)
             dp1 = -q[1] * (1 + 2q[2])
             dp2 = -q[2] - (q[1]^2 - q[2]^2)
-            @SVector [dp1, dp2]
```
SebastianM-C: Why did you remove `SVector` initial conditions?

src/common.jl
```diff
@@ -186,40 +186,28 @@ function DiffEqBase.solve(
...
-        u0 = convert(Array{eltype(prob.u0)}, prob.u0)
-        _prob = ODEProblem(f, u0, prob.tspan, prob.p; prob.kwargs...)
-        # DiffEqBase.solve(prob, _alg, args...; kwargs...)
-        integrator = DiffEqBase.__init(_prob, _alg, args...; kwargs...)
-    else
-        f = prob.f
-        if !isinplace && typeof(prob.u0) <: AbstractArray
-            f! = (du, u, p, t) -> (du .= f(u, p, t); 0)
+    f = prob.f
+    if !isinplace && typeof(prob.u0) <: AbstractArray
```
SebastianM-C, line 190: Is `typeof(prob.u0) <: AbstractArray` necessary?
I think that the only case that we have `prob.u0` not be an array is when it is a scalar differential equation, but in that case it can only be out-of-place, so `!isinplace` is already true.

SebastianM-C (commented):
This is a much better solution as it preserves the problem type and avoids converting the user input to an `Array`. This should improve both performance and generality/composability.

test/common.jl
PerezHz: Thanks for checking this! I removed them temporarily because it doesn't work with latest commit 😅 since `setindex!` is not defined for `SArray{S,Taylor1}`, I'm trying to fix that right now, just pushed this to see how travis reacted

PerezHz: I'll add them again as soon as I fix this

src/common.jl
```diff
@@ -1,194 +1,258 @@
...
-      push!(tstops_internal, tdir_tf)
+    if verbose
+        warned = !isempty(kwargs) && check_keywords(alg, kwargs, warnlist)
+        warned && warn_compat()
+    end
+
+    f = prob.f
+    if !isinplace && typeof(prob.u0) <: AbstractArray
+        if prob.f isa DynamicalODEFunction
+            f1! = (dv, v, u, p, t) -> (dv .= prob.f.f1(v, u, p, t); 0)
```
SebastianM-C, line 192: Since we are in the out-of-place case, shouldn't this _not_ use broadcasting to modify the elements of `dv` in-place? 
```suggestion
            f1! = (dv, v, u, p, t) -> (dv = prob.f.f1(v, u, p, t); 0)
```

src/common.jl
```diff
@@ -1,194 +1,258 @@
...
+            f2! = (du, v, u, p, t) -> (du .= prob.f.f2(v, u, p, t); 0)
```
SebastianM-C, line 193: Since we are in the out-of-place case, shouldn't this _not_ use broadcasting to modify the elements of `dv` in-place?
```suggestion
            f1! = (dv, v, u, p, t) -> (dv = prob.f.f1(v, u, p, t); 0)
            f2! = (du, v, u, p, t) -> (du = prob.f.f2(v, u, p, t); 0)
```

src/common.jl
SebastianM-C, line 192: This is what causes test to fail in the OOP case with immutable types.

SebastianM-C (commented):
The translation of out-of-place problems to in-place ones is fundamentally incompatible with immutable types. From what I understand from the DiffEq code, the OOP code path is made specifically for immutable types.

src/common.jl
PerezHz: Thank you for your suggestion. If we think an oop scalar problem, and if we removed that `typeof(prob.u0) <: AbstractArray`, then in that case the code would try to transform it into a in-place scalar problem, but currently TaylorIntegration does not support in-place scalar problems; that's why I left it there. Tests actually break if I remove that part

test/common.jl
PerezHz: They are back now... not the prettiest solution, but it was the best compromise until we properly support oop problems in TaylorIntegration.jl

src/common.jl
PerezHz: Good catch, thanks! I'm proposing, as a temporary solution for this specific case, to convert the initial condition `prob.u0` from an immutable to a mutable array type (see below)

PerezHz:
Ok, so I think I found a compromise: I found a way to make things to work with `SVectors` when dealing with oop `DynamicalODEProblem` by transforming specifically for this case the initial condition `prob.u0` from an immutable array type to a mutable one (it's highly non-elegant, but does the job). I fully agree that this does not solve the incompatibility of the oop->iip translation with immutable arrays, but solving that deeper issue requires some changes under the hood of TaylorIntegration.jl itself, which I believe we should think through and tackle in another PR.

.gitignore
```diff
@@ -3,3 +3,4 @@ examples/.ipynb_checkpoints
 examples/JuliaCon2017/.ipynb_checkpoints/
 docs/build/
 docs/site/
+.vscode/
```
lbenet, line 6: This may need  a new line (it's the end of the file); this is noted by github.

src/common.jl
```diff
@@ -1,194 +1,267 @@
...
+alg_order, alg_cache, initialize!, perform_step!, @unpack,
+@cache, stepsize_controller!, step_accept_controller!
+
+# TODO: check which keywords work fine
+const warnkeywords = (:save_idxs, :d_discontinuities, :unstable_check, :save_everystep,
+:save_end, :initialize_save, :adaptive, :dt, :reltol, :dtmax,
+:dtmin, :force_dtmin, :internalnorm, :gamma, :beta1, :beta2,
+:qmax, :qmin, :qsteady_min, :qsteady_max, :qoldinit, :failfactor,
+:isoutofdomain, :unstable_check,
+:calck, :progress, :timeseries_steps, :dense)
```
lbenet, line 18: Can you include some indentation; makes nicer reading the code.

Can you open an issue to  have present the TO DO indicated here?

src/common.jl
```diff
@@ -1,194 +1,267 @@
...
+abstract type TaylorAlgorithm <: OrdinaryDiffEqAdaptiveAlgorithm end
 struct TaylorMethod <: TaylorAlgorithm
     order::Int
 end
+struct _TaylorMethod <: TaylorAlgorithm
+    order::Int
+    parse_eqs::Bool
+end
+
+_TaylorMethod(order; parse_eqs=true) = _TaylorMethod(order, parse_eqs) # set `parse_eqs` to `true` by default
```
lbenet, line 33: Naive question: why do we need two (essentially identical) structs, instead  of only one? Perhaps by adjusting some constructors, everything works with only one...

src/common.jl
```diff
@@ -1,194 +1,267 @@
...
+
+alg_order(alg::TaylorMethod) = alg.order
+alg_order(alg::_TaylorMethod) = alg.order
+
+# Regarding `isfsal`, thought about setting isfsal to false, but low order
+# interpolation may be helpful (since Taylor interpolation is memory-intensive),
+# so let's set isfsal to true for now, which is the default, so no method
+# definition needed here, so we in principle should support FSAL
+# isfsal(::TaylorMethod) = true # see discussion above
```
lbenet, line 42: Yet, this line is commented... Is it needed? We can certainly have it there commented, but then clarifying a bit would be very enlightening.

src/common.jl
```diff
@@ -1,194 +1,267 @@
...
+#   uprev::uType
+#   tmp::uType
+#   k::rateType
+#   fsalfirst::rateType
+#   tT::tTType
+#   uT::uTType
+#   duT::uTType
+#   uauxT::uTType
+#   parse_eqs::Bool
+# end) )
```
lbenet, line 80: Do we need this? Seems it is a left-over of something you were testing...

src/common.jl
```diff
@@ -1,194 +1,267 @@
...
+    if length(k)<2 || always_calc_begin
+        if typeof(cache) <: OrdinaryDiffEqMutableCache
+        rtmp = similar(u, eltype(eltype(k)))
+        f(rtmp,uprev,p,t)
+        copyat_or_push!(k,1,rtmp)
+        f(rtmp,u,p,t+dt)
+        copyat_or_push!(k,2,rtmp)
+        else
+        copyat_or_push!(k,1,f(uprev,p,t))
+        copyat_or_push!(k,2,f(u,p,t+dt))
```
lbenet, line 249: Can you include some indentation here?

src/common.jl
```diff
@@ -1,194 +1,267 @@
...
+function evaluate!(x::AbstractArray{Taylor1{T},N}, δt::S,
+        x0::AbstractArray{T,N}) where {T<:Number, S<:Number, N}
 
-    tstops_internal, saveat_internal, d_discontinuities_internal
+    # @assert length(x) == length(x0)
+    @inbounds for i in eachindex(x, x0)
+        x0[i] = evaluate( x[i], δt )
+    end
+    nothing
 end
```
lbenet, line 267: Why not  adding it directly to TaylorSeries?

test/common.jl
```diff
@@ -1,16 +1,24 @@
-using TaylorIntegration, Test, DiffEqBase
+using Test, DiffEqBase, TaylorIntegration
```
lbenet, line 1: Does the order matter here? If so, i think it is worth adding  a couple of sentences to the docs.

src/common.jl
```diff
@@ -1,194 +1,267 @@
...
+        _prob = prob
+    end
 
-      for t in d_discontinuities
-        tdir_t = tdir * t
-        tdir_t0 < tdir_t ≤ tdir_tf && push!(tstops_internal, tdir_t)
-      end
+    # DiffEqBase.solve(prob, _alg, args...; kwargs...)
+    integrator = DiffEqBase.__init(_prob, _alg, args...; kwargs...)
+    integrator.dt = stepsize(integrator.cache.uT, integrator.opts.abstol) # override handle_dt! setting of initial dt
```
lbenet, line 220: Previously, there was a default value set for `abstol`; I cannot see it now. So the question is if the user must now to specify it, or is it set somewhere else?

lbenet (approved):
Thanks for the tremendous work  that there is in this PR! I have left mostly trivial remarks  to clarify a couple of points. 

LGTM and I am in favor of merging this, though maybe we may wait for @SebastianM-C's opinion.

.gitignore
PerezHz: Fixed, thanks!

src/common.jl
PerezHz: Thanks, added indentation and also opened the issue corresponding to this TODO

src/common.jl
PerezHz: Good catch! Yes, it was redundant to have those two structures, so I removed one of them

src/common.jl
PerezHz: It was more of a personal reminder when trying to figure the internals of DiffEqBase and OrdinaryDiffEq; I've now removed these commented lines

src/common.jl
PerezHz: 😅 sorry, thanks! Yes, this was a leftover, it's now gone

src/common.jl
PerezHz: Thanks, done!

src/common.jl
PerezHz: This is the only point which I haven't addressed yet; I will open the corresponding PR in TaylorSeries, and it will perhaps require a new TaylorSeries release

test/common.jl
PerezHz: The order makes no difference, I just forgot to leave that line untouched. I've now gone back to the original order, so that this line doesn't show up when doing diffs in the future

src/common.jl
PerezHz: Good catch! If the user doesn't specify `abstol` it is set automatically by `DiffEqBase.__init` to `abstol=1e-6`

PerezHz:
> Thanks for the tremendous work that there is in this PR! I have left mostly trivial remarks to clarify a couple of points.

Thanks @lbenet for your review! I addressed all your comments except one, which will have to wait a bit while we add the new `evaluate!`method to TaylorSeries. In the meantime, please let me know if there's something that you think can be further improved.
 
> LGTM and I am in favor of merging this, though maybe we may wait for @SebastianM-C's opinion.

I agree we should wait to hear Sebastian's opinion.

src/common.jl
PerezHz: Update: just opened the corresponding PR. See: https://github.com/JuliaDiff/TaylorSeries.jl/pull/252

SebastianM-C (approved):

lbenet:
Tests are passing, except in nightly; I am just including the (first) problem spotted by travis [here](https://travis-ci.org/github/PerezHz/TaylorIntegration.jl/jobs/747480423#L598) to address it later.

lbenet:
Shall I go ahead and merge?

PerezHz:
> Shall I go ahead and merge?

Sorry for the late reply, sure, I agree that we should merge!

lbenet:
Merged! Thanks a lot!

PR #118, opened 2020-12-09
Enable GitHub Actions-based CI and docs deployment

PerezHz:


lbenet:
It seems that github actions are way faster than appveyor/travis. Have you considered including windows tests in ci.yaml? (I am having problems to make tests pass in windows; I simply do not understand the problem.)

PerezHz:
> It seems that github actions are way faster than appveyor/travis. Have you considered including windows tests in ci.yaml? (I am having problems to make tests pass in windows; I simply do not understand the problem.)

Thanks! Just updated things to see how maOS and windows CI work; let's hope they work well here

PerezHz:
windows CI seems to be working fine; I'm using a slightly different version of `ci.yml` wrt the one in TaylorSeries.

PerezHz:
@lbenet it looks like the docs are being built and deployed correctly as well, so this is now ready for review

PerezHz:
Just updated the build badge in README.md and fixed coveralls (it's now triggered from Github CI); but this is now really ready for review 😅

.github/workflows/TagBot.yml
```diff
@@ -1,11 +1,15 @@
...
+  workflow_dispatch:
 jobs:
   TagBot:
+    if: github.event_name == 'workflow_dispatch' || github.actor == 'JuliaTagBot'
     runs-on: ubuntu-latest
     steps:
       - uses: JuliaRegistries/TagBot@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }}
```
lbenet, line 15: Shouldn't this read: `ssh: ${{ secrets.DOCUMENTER_KEY }}`?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: '1'
+      - name: Install dependencies
+        run: julia --project=docs/ -e 'using Pkg;
+                Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
+      - name: Build and deploy
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For authentication with GitHub Actions token
+          DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }} # For authentication with SSH deploy key
```
lbenet, line 27: Same comment as above

lbenet (approved):
Thanks a lot! 

I left two comments, which honestly, I do not know if they should be  as i propose or not.  Tests are passing, so just  let  me know if you plan to change something  or  if I should just  merge it. I'm in favor of merging.

.github/workflows/TagBot.yml
PerezHz: Good catch, thanks! It's now fixed

.github/workflows/docs.yml
PerezHz: According to Documenter.jl [docs](https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#GitHub-Actions) it seems that in this case it should be fine as it is now. Maybe we can try to see how it works, and fix it if necessary

lbenet: Just check, and  this comment does not apply... Sorry.

lbenet:
I think once the tests pass, I'll merge. Thanks a lot once again!

PerezHz:
> I left two comments, which honestly, I do not know if they should be as i propose or not. Tests are passing, so just let me know if you plan to change something or if I should just merge it. I'm in favor of merging.

I changed TagBot.yml as you propose, and left the other `DOCUMENTER_KEY` occurence as it is, since those are the conventions used in Documenter.jl. Otherwise, I agree we should merge this.

> I think once the tests pass, I'll merge.

Great!

> Thanks a lot once again!

No problem!

PR #123, opened 2021-01-29
Create CompatHelper.yml

mforets:
This adds the CompatHelper bot. However, it rmay remain to setup deploy keys (https://github.com/JuliaRegistries/CompatHelper.jl#1-installation), and I don't have privileges in this repo to help with that matter.

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -1,0 +1,16 @@
+name: CompatHelper
+on:
+  schedule:
+    - cron: '00 00 * * *'
+  workflow_dispatch:
+jobs:
+  CompatHelper:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Pkg.add("CompatHelper")
+        run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
+      - name: CompatHelper.main()
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMPATHELPER_PRIV: ${{ secrets.COMPATHELPER_PRIV }}  # optional
+        run: julia -e 'using CompatHelper; CompatHelper.main()'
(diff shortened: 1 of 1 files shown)


.github/workflows/CompatHelper.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  CompatHelper:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Pkg.add("CompatHelper")
+        run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
+      - name: CompatHelper.main()
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMPATHELPER_PRIV: ${{ secrets.COMPATHELPER_PRIV }}  # optional
```
lbenet, line 15: Do we have to define `COMPATHELPER_PRIV` token or not? If so, could we use `DOCUMENTER_KEY` as in [here](https://github.com/PerezHz/TaylorIntegration.jl/blob/76cd51a807f27bd7a9a51fc96e4a088884f83248/.github/workflows/docs.yml#L27)?

lbenet (commented):
Thanks a lot! I left one comment due to (my) ignorance. Aside from that, LGTM!

.github/workflows/CompatHelper.yml
mforets: indeed, im no expert but i think you can also remove it 
 see other packages eg .https://github.com/SciML/DifferentialEquations.jl/blob/master/.github/workflows/CompatHelper.yml

lbenet: Can you comment it, so if we require it it is a simple change? Then I merge this!

mforets, line 15:
```suggestion
```

lbenet (approved):
Thanks a lot! I merge this once the tests are finished!

lbenet:
Thanks again @mforets. Merging!

PR #126, opened 2021-02-06
Update restricted three-body problem example

### LINKED ISSUE #125: Clarification: Performance of Vern9
Hello, very interesting package!

I just looked through the docs and found that you are comparing taylorintegration to `Vern9`. I was surprised to see that this higher order explicit method fails to visit the second mass and noticed that you are calling the solver with

`solV = solve(prob, Vern9(), abstol=1e-20); #solve `prob` with the `Vern9` method`
 
that is, the `reltol` value is not set, in which case it should default to `reltol=1e-3` (see [here](https://diffeq.sciml.ai/stable/basics/common_solver_opts/)), at least this is what happens for lower order methods. Do you get the same qualitative behaviour for 

`solV = solve(prob, Vern9(), reltol = 1e-20, abstol=1e-20);`?

PerezHz:
Fixes #125. Thanks to @lindnemi for the heads up! Incidentally, I found and fixed a bug which caused `parse_eqs` turn to `false` always when using the common interface.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -36,7 +36,6 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 TaylorSeries = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
-BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 
 [targets]
-test = ["TaylorSeries", "LinearAlgebra", "Test", "Elliptic", "InteractiveUtils", "Pkg", "StaticArrays", "BenchmarkTools"]
+test = ["TaylorSeries", "LinearAlgebra", "Test", "Elliptic", "InteractiveUtils", "Pkg", "StaticArrays"]
--- a/Project.toml
+++ b/Project.toml
@@ -36,6 +36,7 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 TaylorSeries = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 
 [targets]
-test = ["TaylorSeries", "LinearAlgebra", "Test", "Elliptic", "InteractiveUtils", "Pkg", "StaticArrays"]
+test = ["TaylorSeries", "LinearAlgebra", "Test", "Elliptic", "InteractiveUtils", "Pkg", "StaticArrays", "BenchmarkTools"]
--- a/docs/src/common.md
+++ b/docs/src/common.md
@@ -75,11 +75,13 @@ The equations of motion for the PCR3BP are
 \end{aligned}
 ```
 
-We define this system of ODEs in a way that allows the use of the [`@taylorize`](@ref taylorize)
-macro from `TaylorIntegration.jl`, which under certain circumstances allows
-important speed-ups.
+We define this system of ODEs in a way that allows the use of the [`@taylorize`](@ref)
+macro from `TaylorIntegration.jl`, which for the present example allows
+important speed-ups. For more details about the specifics of the use of [`@taylorize`](@ref),
+see [this section](@ref taylorize).
 ```@example common
-function pcr3bp!(dq, q, param, t)
+using TaylorIntegration
+@taylorize function pcr3bp!(dq, q, param, t)
     local μ = param[1]
     local onemμ = 1 - μ
     x1 = q[1]-μ
@@ -112,7 +114,7 @@ zero-velocity curves on the ``x``-axis.
 ```@example common
 ZVC(x) =  -x^2/2 + V(x, zero(x)) # projection of the zero-velocity curves on the x-axis
 
-plot(ZVC, -2:0.001:2, label="zero-vel. curve", legend=:topleft)
+plot(ZVC, -2:0.001:2, label="zero-vel. curve", legend=:topleft, fmt = :png)
 plot!([-2, 2], [-1.58, -1.58], label="J0 = -1.58")
 ylims!(-1.7, -1.45)
 xlabel!("x")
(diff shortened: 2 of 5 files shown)


Project.toml
```diff
@@ -36,6 +36,7 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 TaylorSeries = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 
 [targets]
-test = ["TaylorSeries", "LinearAlgebra", "Test", "Elliptic", "InteractiveUtils", "Pkg", "StaticArrays"]
+test = ["TaylorSeries", "LinearAlgebra", "Test", "Elliptic", "InteractiveUtils", "Pkg", "StaticArrays", "BenchmarkTools"]
```
lbenet, line 42: I'm not sure that we require to have `BenchmarkTools` here...

docs/src/common.md
```diff
@@ -75,26 +75,31 @@ The equations of motion for the PCR3BP are
...
+    y = q[2]
+    ysq = y^2
+    r1_1p5 = (x1sq+ysq)^1.5
+    x2 = q[1]+onemμ
+    x2sq = x2^2
+    r2_1p5 = (x2sq+ysq)^1.5
+    dq[1] = q[3] + q[2]
+    dq[2] = q[4] - q[1]
+    dq[3] = -((onemμ*x1)/r1_1p5) - ((μ*x2)/r2_1p5) + q[4]
+    dq[4] = -((onemμ*y )/r1_1p5) - ((μ*y )/r2_1p5) - q[3]
```
lbenet, line 98: Not sure if it helps for performance, but i would add one (pair) parenthesis to truly make this evaluation a composition of binary operations, so that it reads as `(a+b)+c` or `a + (b+c)`.

docs/src/common.md
```diff
@@ -151,36 +156,41 @@ H(q0) == J0
...
+via its common interface bindings with `DiffEqBase.jl`; both packages need to
+be loaded explicitly.
 ```@example common
-tspan = (0.0, 1000.0)
+tspan = (0.0, 2000.0)
 p = [μ]
 
-using TaylorIntegration, DiffEqBase
-prob = ODEProblem(f, q0, tspan, p)
+using DiffEqBase
```
lbenet, line 166: While it doesn't do anything here, I would still load both packages for consistency to what  is written.

docs/src/common.md
```diff
@@ -151,36 +156,41 @@ H(q0) == J0
...
+We solve `prob` using a 25-th order Taylor method, with a local absolute tolerance ``\epsilon_\mathrm{tol} = 10^{-15}``.
 ```@example common
-solT = solve(prob, TaylorMethod(25), abstol=1e-20);
+solT = solve(prob, TaylorMethod(25), abstol=1e-15);
 ```
 
 As mentioned above, we load `OrdinaryDiffEq` in order to solve the same problem `prob`
-now with the `Vern9` method, which the `DifferentialEquations.jl`
+now with the `Vern9` method, which the `DifferentialEquations`
 [documentation](http://docs.juliadiffeq.org/latest/solvers/ode_solve.html#Non-Stiff-Problems-1)
```
lbenet, line 177: The link to DifferentialEquations.jl docs seems broken; I think the correct one is https://diffeq.sciml.ai/stable/solvers/ode_solve/#Non-Stiff-Problems

src/common.jl
```diff
@@ -230,3 +230,15 @@ function DiffEqBase.addsteps!(k, t, uprev, u, dt, f, p, cache::TaylorMethodCache
     update_jetcoeffs_cache!(u,f,p,cache)
     nothing
 end
+
+@inline __jetcoeffs!(::Val{false}, f::ODEFunction, t, x, params) =
+    jetcoeffs!(f.f, t, x, params)
+@inline __jetcoeffs!(::Val{true},  f::ODEFunction, t, x, params) =
+    jetcoeffs!(Val(f.f), t, x, params)
```
lbenet, line 237: It seems that this line is not tested...

docs/src/common.md
```diff
@@ -151,36 +156,41 @@ H(q0) == J0
 
 Following the `DifferentialEquations.jl`
 [tutorial](http://docs.juliadiffeq.org/latest/tutorials/ode_example.html),
-we define an `ODEProblem` for the integration;
-`TaylorIntegration.jl` can be used via its common interface bindings with `DiffEqBase.jl`; both packages need to be loaded explicitly.
+we define an `ODEProblem` for the integration; `TaylorIntegration` can be used
```
lbenet, line 159: This link is broken.

docs/src/common.md
```diff
@@ -194,19 +204,15 @@ For comparison, we now plot the orbit corresponding to the
...
 ylabel!("y")
 ```
-We note that the orbits do not display the same qualitative features. In particular,
-the `Vern9()` integration displays an orbit which does not visit the secondary,
-as it was the case in the integration using Taylor's method, and stays far
-enough from ``m_1``. The question is which integration should we trust?
-
-We can obtain a quantitative comparison of the validity of both integrations
+We note that both orbits display the same qualitative features. We can obtain a
+quantitative comparison of the validity of both integrations
```
lbenet, line 215: I agree that both orbits display the same qualitative properties, but the plots show also qualitative differences, which I  think it  should also be added. For instance, the integration with our package gets closer to the primary  than that with Verrn9.

docs/src/common.md
```diff
@@ -218,57 +224,62 @@ nothing # hide
...
 xlabel!("t")
 ylabel!("dE")
 ```
-We notice that the Jacobi constant absolute error for the `TaylorIntegration.jl`
-solution remains bounded below ``5\times 10^{-14}``, despite of the fact that
-the solution displays many close approaches with ``m_2``.
+We notice that the Jacobi constant absolute error for the `TaylorMethod(25)`
+solution remains bounded below ``2\times 10^{-14}``, while the `Vern9()` solution
+displays sudden jumps earlier in the integration and has variations larger than
+``10^{-13}``.
```
lbenet, line 248: Our solution display changes in the Jacobi constant below 10^(-13); the Vern9 solution is, at the end of the integration time, two order of magnitude above our solution for the same quantity. This  is also worth mentioning.

lbenet (commented):
Thanks for this PR. It  truly looks good to me!

I have submitted some comments just to emphasize some aspects of the comparison of the results. Maybe the most important comments are related to some broken links, which should point to the DiffEqs docs. I checked this reading directly the docs [here](https://perezhz.github.io/TaylorIntegration.jl/previews/PR126/common/).

Project.toml
PerezHz: Thanks, fixed!

docs/src/common.md
PerezHz: Parentheses added 👍

docs/src/common.md
PerezHz: Ok, re-added TaylorIntegration there

docs/src/common.md
PerezHz: Thanks, fixed!

src/common.jl
PerezHz: Thanks, added a test for this line

docs/src/common.md
PerezHz: Link updated

docs/src/common.md
PerezHz: Ok, updated discussion

docs/src/common.md
PerezHz: Sure, updated!

PerezHz:
Thank you for your review, @lbenet! I updated the docs and the tests; let me know if there's anything else I missed

PerezHz:
Also, since merging this PR fixes a bug, should we bump the patch version?

lbenet (approved):
LGTM! Please go ahead and merge this. I also agree to release a new patch version. thanks a lot again!

PR #131, opened 2021-06-10
Get rid of errors in Poincare section docs

lbenet:


lbenet:
I open this PR because I noticed that the *current* docs, in the section on [Poincare maps](https://perezhz.github.io/TaylorIntegration.jl/stable/root_finding/) displays some odd errors, just before the animations are created. Locally, after removing some lines, I got rid of the error messages, though they appeared on the terminal; yet, the quality of the animation is worst.

lbenet:
See the result [here](https://perezhz.github.io/TaylorIntegration.jl/previews/PR131/root_finding/).

lbenet:
Somehow, the produced gif has worst resolution that the one loaded...

PerezHz:
The last commit seems to correctly get rid of the errors? The problem seems solved, would you like me to merge this, or is there anything else to be added here?

PerezHz:
Thanks for looking into this!

lbenet:
> The last commit seems to correctly get rid of the errors? The problem seems solved, would you like me to merge this, or is there anything else to be added here?

The solution is good enough, I think. So yes, go ahead and merge it!

PerezHz:
Merged! Again, thanks!

Issue #140, opened 2022-03-13
Conditional output size and @taylorize?


vram0gh:
I have an ODE system that can include or exclude an equation depending on how the the solution is being used.

We know input vector length is not a part of the type signature without using something like `StaticArrays.jl`.  So for `DifferentialEquations.jl` ODEs, I check the length of the input vector to determine if the additional component is included.  Within an `if` block, I compute the augmenting ODE RHS if the input vector contains that dependent variable.

Currently with `@taylorize`, wrapping the computation in an `if` guard in the original function definition doesn't prevent the generated code from always attempting to access the additional input element, which may not exist.

What is the best approach for handling this situation when using `TaylorIntegration.jl`?  The options seem to be:

- Have two ODE functions with different names: one with the additional equation, one without.  If I do this and still want both functions to be optimized by `@taylorize`, what is the proper way to avoid literal duplication of the common computations into both functions, or overhead from unnecessary copies/allocations/function calls?
- Use `StaticArrays.jl` to have two different methods for the same name, one with the extra component, one without.  If going this route, again, what is the best way to share the common code in a way that works well with `@taylorize`?
- Feature request that `@taylorize` recognize the conditional guard on the computation and output of additional variables.

Am I missing an obvious or easy option?  What is the recommended approach?  Apologies if opening this issue is not the right way to go about this.

lbenet:
Thanks for opening this. I would have (naively) though that it would work with an `if` statement of the form:
```julia
function myeqs(du, u, params, t)
    # Some code 
    if length(du) == 5 # 5 is an arbitrary example
        du[5] = ...
    end
    return du
end
```

I can't check this right now, but I will try to get back to it later. If you can provide us with a minimum (non-)working example, this could speed things up...

vram0gh:
Thank you; here is a simple example:

```julia
@taylorize function var_simpletest!(primes, dependent, params, independent)
  x = independent
  y = dependent[1]
  primes[1] = ((y * sin(x))^2) / (x * y)
  if (length(dependent) > 1)
    r = dependent[2]
    primes[2] = r * primes[1]
  end
  return nothing
end

function Taylor_poly_var_simpletest()
  ord = 20
  x = Taylor1(Float64, ord)
  x[0] = 1.0
  y = [Taylor1(2.0, ord)]
  yprime_out = [Taylor1(0.0, ord)]
[...]

PR #141, opened 2022-03-14
Document @taylorize limitations discussed in #140 and lightly edit language

vram0gh:
These are edits to `docs/src/taylorize.md`.  The first commit directly addresses (#140) and the second contains light edits to improve the language (make idiomatic, fix small grammar issues).  Maintainer input & iteration welcome.

### DIFF
--- a/docs/src/taylorize.md
+++ b/docs/src/taylorize.md
@@ -1,18 +1,18 @@
 # [Optimizing: `@taylorize`](@id taylorize)
 
 Here, we describe the use of the macro [`@taylorize`](@ref), which
-parses the functions containing the ODEs to be integrated, allowing to speed up
-[`taylorinteg`](@ref) and [`lyap_taylorinteg`](@ref).
+parses the functions containing the ODEs to be integrated, allowing
+[`taylorinteg`](@ref) and [`lyap_taylorinteg`](@ref) to be sped up.
 
 !!! warning
-    The macro [`@taylorize`](@ref) is still in an experimental phase;
+    The macro [`@taylorize`](@ref) is still in an experimental state;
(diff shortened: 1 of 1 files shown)


docs/src/taylorize.md
```diff
@@ -1,18 +1,18 @@
 # [Optimizing: `@taylorize`](@id taylorize)
 
 Here, we describe the use of the macro [`@taylorize`](@ref), which
-parses the functions containing the ODEs to be integrated, allowing to speed up
-[`taylorinteg`](@ref) and [`lyap_taylorinteg`](@ref).
+parses the functions containing the ODEs to be integrated, allowing
+[`taylorinteg`](@ref) and [`lyap_taylorinteg`](@ref) to be sped up.
```
lbenet, line 5: I think this should read "to be speed up".

lbenet (commented):
Thanks a lot for improving this part of the documentation. I left a single comment (a typo); aside from that, LGTM.

docs/src/taylorize.md
vram0gh: My pleasure, this is the least I can do.

On this line, it could be either "sped" or "speeded" as in the intransitive form of speed.  While not authoritative, you can see from https://english.stackexchange.com/questions/114620/speeded-vs-sped how uncommon it is to use "speeded".  Anecdotally, my web browser is highlighting a supposed spelling error in "speeded", indicating that it's so uncommon that it's not in the dictionary the browser is using.

lbenet: I stand corrected... I checked against a dictionary, and [sped](https://www.merriam-webster.com/dictionary/SpEd) is ok :-)

I'll merge this.

PR #142, opened 2022-04-01
Metaprogramming for taylorinteg dense output

LuEdRaMo:
Metaprogramming for `taylorinteg` dense output methods in order to avoid code repetition.

### DIFF
--- a/src/explicitode.jl
+++ b/src/explicitode.jl
@@ -302,8 +302,71 @@ end
 
 
 # taylorinteg
+for V in (:(Val{true}), :(Val{false}))
+    @eval begin
+        function taylorinteg(f, x0::U, t0::T, tmax::T, order::Int, abstol::T, ::$V, params = nothing;
+            maxsteps::Int=500, parse_eqs::Bool=true) where {T<:Real, U<:Number}
+    
+            # Allocation
+            tv = Array{T}(undef, maxsteps+1)
+            xv = Array{U}(undef, maxsteps+1)
+            if $V == Val{true}
+                polynV = Array{Taylor1{U}}(undef, maxsteps+1)
+            end
+
+            # Initialize the Taylor1 expansions
+            t = Taylor1( T, order )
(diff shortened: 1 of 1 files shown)


lbenet:
Thanks @LuEdRaMo for your contribution. I just allowed running the tests. Sorry for the delay...

src/explicitode.jl
```diff
@@ -350,216 +411,89 @@ tv, xv, polynv = taylorinteg(f!, [3, 3], 0.0, 0.3, 25, 1.0e-20, maxsteps=100, Va
...
-            """)
-            break
-        end
-    end
-
-    #return tv, xv
-    return view(tv,1:nsteps), view(xv,1:nsteps)
+function taylorinteg(f, x0::U, t0::T, tmax::T, order::Int, abstol::T, params = nothing;
+    maxsteps::Int=500, parse_eqs::Bool=true) where {T<:Real, U<:Number}
+    return taylorinteg(f, x0, t0, tmax, order, abstol, Val(false), params; maxsteps, parse_eqs)
```
lbenet, line 416: I think the errors that appear when testing with Julia v1.0 (for example [here](https://github.com/PerezHz/TaylorIntegration.jl/runs/5821366447?check_suite_focus=true#step:6:27)) are due to the fact that you must include the keyword explicitly for the keyword arguments of a function. Can you test this?

lbenet (commented):
I left one comment which applies to two parts of the code and is related to the failures in the tests for Julia v1.0. Aside from that, it LGTM! Thanks a lot!

Once said this, I was wondering if we should merge #137 before this one, since both PRs modify `src/explicitode.jl`, and more files would be needed to be readapted in #137 if we merge this one before.

lbenet (approved):
Tests pass and you have taken into account my comments. So I think this is good to be merged! Thanks a lot!

lbenet:
Merged!

Issue #149, opened 2022-08-22
`ODEProblem`s with `recompile=false` flag not supported


nathanaelbosch:
One way to reduce additional recompilations is to construct the ODE problem with `ODEProblem{true,false}(...)`. This is currently not supported by TaylorIntegration.jl, but it would be great to have this.

Minimal example:
```julia
using OrdinaryDiffEq, TaylorIntegration
function lorenz(du, u, p, t)
    du[1] = 10.0(u[2] - u[1])
    du[2] = u[1] * (28.0 - u[3]) - u[2]
    du[3] = u[1] * u[2] - (8 / 3) * u[3]
    return nothing
end
prob = ODEProblem{true,false}(lorenz, [1.0; 0.0; 0.0], (0.0, 1.0))
solve(prob, TaylorMethod(25))
```

I also observed that `TaylorIntegration.jetcoeffs!` fails when `f` comes from the problem definitions above. 

EDIT: Fixed typos in the minimal example and added "using OrdinaryDiffEq" for completeness.

lbenet:
Hi! Thanks for drawing our attention to this problem. 

I think it is fair to say that we are a bit behind in everything related to provide a fully functional common interface with DifferentialEquations.jl. Today it is a bit complicated for me to have a look on this, but I will try to find some time the next few days.

Once said this, I think I should mention [`@taylorize`](https://perezhz.github.io/TaylorIntegration.jl/latest/taylorize/) as a way to optimize allocations and runtime in the code, which somewhat deals with the peculiarities of Taylor's integration method.

nathanaelbosch:
I see, thanks for the reply. To be fair, I do not actually ever use the DifferentialEquations.jl interface, I only came by the issue because I rely on `jetcoeffs!` to initialize the ODE solvers in ProbNumDiffEq.jl ([code here](https://github.com/nathanaelbosch/ProbNumDiffEq.jl/blob/42e58bba71af26ce0d7ce90a8da2e32def42cadf/src/initialization/taylormode.jl#L31-L45)), and following some recent discussion about reducing compilations by using `ODEProblem{true,false}` I wanted to get similar results in ProbNumDiffEq.jl. In any case, thanks a lot for having a look!

lbenet:
I'll try to have a look on your use case during this weeks. If you can provide a link to the discussion you mention about using `ODEProblem{true,false}`, that will be helpful!

nathanaelbosch:
[...]

PR #150, opened 2022-10-24
Update DiffEq interface methods

PerezHz:
Following the discussion and the suggestions in #149, this PR adds minor updates to the common DiffEq interface methods. Specifically, I added a `SciMLBase.unwrapped_f(prob.f)` at the top of the `solve` method overloading for `TaylorMethod`, together with other minor fixes. I'm also proposing to turn off testing on Julia 1.0 and keep testing on the new LTS version (Julia 1.6), together with the current stable release and nightly. cc: @lbenet


src/common.jl
```diff
@@ -1,4 +1,4 @@
-using DiffEqBase, OrdinaryDiffEq
+using DiffEqBase, OrdinaryDiffEq, SciMLBase
```
lbenet, line 1: Naive question: is it possible to reduce the dependencies here? Or each one is truly needed?

lbenet (approved):
Thanks a lot @PerezHz. I

 only left one question related to the number of dependencies; it would be great if we could somehow reduce them, though I understand that "a must is a must". 

LGTM! Feel free to merge it and tag a new minor(?) version

src/common.jl
PerezHz: Good catch! I hadn't noticed that DiffEqBase re-exports SciMLBase; so it seems it's not necessary to write down the explicit dependency. We only have to be careful to have the right SciML Base version (at least 1.51), but I think we already have more than enough (1.65). I pushed a new commit which addresses this.

PR #151, opened 2022-11-08
Add benchmarking infrastructure

lbenet:
This PR (draft) adds some needed infrastructure to produce benchmarks, using [PkgBenchmark.jl](https://github.com/JuliaCI/PkgBenchmark.jl) and [BenchmarkTools.jl](https://github.com/JuliaCI/BenchmarkTools.jl). The idea of opening this *work-in-progress* is to discuss and perhaps modify the design than the one I have outlined here. This is based in something similar done in TaylorModels.jl. 

The actual motivation is an observation of @LuEdRaMo, who found a concrete case where using the new (current) version of `@taylorize` yields a regression in performance. In order then to check `@taylorize` of v0.8.11 against the current version (v0.10.0 as for today) I have created a fake 0.8.12 version in the branch [test/v0.8.12](https://github.com/PerezHz/TaylorIntegration.jl/tree/test/v0.8.12) which has the same files added here (which is somehow required by `PkgBenchmark.jl`). In order to run the benchmarks added, you need simply to run the script `run_benchmarks..jl` located at the main directory of the repo.

At the moment, I'm not sure that `PkgBenchmark.jl` is actually a required dependency, but it is confortable to create reports with the timing and memory comparisons.


lbenet:
I forgot to mention that in principle it is possible to automatize the creation of the benchmarks; see [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark/blob/master/.github/workflows/julia.yml).

lbenet:
I've added a new benchmark, which is realetd with aux array `__arrTaylor::Vector{Vector{Taylor1}}`, which should be avoiding repeated allocations when an array of `Taylor1` objects is declared within the ODE declaring function. 

The results are not positive, so indeed in this case, there is a regression when such array is needed.

For the time being, I think it is worth to review this PR, so new improvements can use it.

cc @PerezHz @LuEdRaMo

lbenet:
I think we should go ahead and merge this, to allow using it (though not yet automatically) in other PRs. Shall I go ahead @PerezHz ?

benchmark/manyspin_benchmarks.jl
```diff
@@ -0,0 +1,111 @@
...
+        q[3i-2] = as * cos(pha)
+        q[3i-1] = as * sin(pha)
+        q[3i  ] = sz[i]
+    end
+
+    return nothing
+end
+
+const _abstol = 1.0e-20
+const _order = 26
```
PerezHz, line 82: Maybe it makes sense to move the repeated constants (here and in `kepler_benchmarks.jl` to `benchmarks.jl`?

PerezHz (approved):
Many thanks @lbenet for this! This is really helpful! I just left a single comment regarding repeated constants in the benchmarks. Other than that, LGTM!

benchmark/manyspin_benchmarks.jl
PerezHz: By repeated constants I mean `_order`, `tf1` and `tf2`, which are declared both in `manyspin_becnhmarks.jl` and `kepler_benchmarks.jl`

lbenet: I have kept all constants within the corresponding benchmark file, since maybe we want to change them depending on the actual benchmark. I put them, together with the actual `@benchmarkable` calls, within a `let ... end` block, which avoids them to get some warnings.

I've also moved `run_benchmarks.jl` file to the `benchmark` dir, and updated the README.md file there.

PerezHz: Okay, makes a lot of sense!

PerezHz:
Got the green lights from GHA, so went ahead and merged! Many thanks, @lbenet!

Issue #153, opened 2022-11-26
UndefVarError occurs when calling _allocate_jetcoeffs!


vram0gh:
I just restarted use of `TaylorIntegration.jl` following #137 becoming a part of the released version.  I'm not sure if the below is a bug in the current `@taylorize` or a misuse on my part:

```julia
using TaylorIntegration

@taylorize function odesys_test!(primes, dependent, params, independent)
    local p1 = params[1]
    local p1m1 = (p1 - 1)
    local u1_scalefactor = params[2]
    local u2_scalefactor = params[3]
    θ = independent
    u1 = dependent[1] * u1_scalefactor
    u2 = dependent[2] * u2_scalefactor
    u1sq = u1 ^ 2
    u2sq = u2 ^ 2
    T = (1 - u1sq) - u2sq
    primes[1] = u2 / u1_scalefactor
    primes[2] = (((u1 * u2sq) - (p1m1 * ((2 * u1) + (u2 * cot(θ))) * T)) / ((p1m1 * T) - u2sq)) / u2_scalefactor
  end
```

REPL output, Julia 1.8.2, TaylorIntegration.jl 0.10.0
```julia
julia> order = 10
julia> t_start = 1.75
julia> t_Taylor = Taylor1(t_start, order)
julia> IVs = [0.18, 2.1]
julia> dependent_Taylor=Taylor1.(IVs, order)
julia> primes_Taylor = deepcopy(dependent_Taylor)
julia> TaylorIntegration._allocate_jetcoeffs!(Val(odesys_test!), t_Taylor, dependent_Taylor, primes_Taylor, [1.2, 1.0, 1.0])
ERROR: UndefVarError: tmp789 not defined
Stacktrace:
 [1] _allocate_jetcoeffs!(#unused#::Val{odesys_test!}, independent::Taylor1{Float64}, dependent::Vector{Taylor1{Float64}}, primes::Vector{Taylor1{Float64}}, params::Vector{Float64})
   @ Main ./none:0
 [2] top-level scope
   @ REPL[32]:1
```

My use case is having TaylorIntegration.jl generate the jet of coefficients, but using these coefficients for other purposes.  With the prior version of `_jetcoeffs!`, I was able to accomplish this without issue.

`_allocate_jetcoeffs!` does work for a simpler example:
```julia
@taylorize function odesys_test2!(primes, dependent, params, independent)
  local p1 = params[1]
  local p2 = params[2]
  local p3 = params[3]
[...]

PR #154, opened 2022-11-27
Some proposed @taylorize doc updates based on experience from Issue #153

vram0gh:
Make clear the two operand expression rule isn't just a good practice, but required. Also add an explicit note about debugging the generated code. Follows from Issue PerezHz#153.


lbenet:
Thanks a lot! LGTM, so I'll go ahead and merge!

PR #158, opened 2023-02-27
Benchmarks for Lyapunov exponents

AdrianMI49:


benchmark/Lyap_benchmark.jl
```diff
@@ -0,0 +1,37 @@
+#Lyapunov Benchmarks
+
+@taylorize function lorenz!(dx,x,param,t)
+    σ, ρ, β = param
+     dx[1] = σ*(x[2]-x[1])
```
lbenet, line 5: Comment on style: please delete one space before `dx[1]`.

benchmark/Lyap_benchmark.jl
```diff
@@ -0,0 +1,37 @@
...
+    const _order = 28
+    const pars = [16.0, 45.92, 4.0]
+    const t0 = 0.0
+    const tf1 = 1000.0
+    const maxsteps1 = 500_000
+    const tf2 = 10_000.0
+    const maxsteps2 = 50_000_000
+    const tf3 = 100_000.0
+    const maxsteps3 = 500_000_000
+    q0 = [0.2, 0.0, 0.0, 3.0]
```
lbenet, line 24: Three comments in these lines:
- Please delete the 4 spaces (or tab), so everything is properly aligned
- `q0` is the initial condition, so it must have the same dimension as the equations of motion, i.e. 3. I'd suggest to use `q0 = [19.0, 20.0, 50.0]`, as in the tests. Also, perhaps having it as `const` is ok.
- You need to add the line `set_variables("δ", order=1, numvars=3);", so the stability matrix can be computed using automatic differentiation.

benchmark/Lyap_benchmark.jl
```diff
@@ -0,0 +1,37 @@
...
+
+
+SUITE["Lorenzlyap"] = BenchmarkGroup()
+    
+SUITE["Lorenzlyap"]["lorenzlyap1-1"] = @benchmarkable taylorinteg(
+        lorenz!, $q0, $t0, $tf1, $_order, $_abstol, #=Val(true)  o Val(false),=# $pars, -1.0, maxsteps=$maxsteps1)
+SUITE["Lorenzlyap"]["lorenzlyap1-2"] = @benchmarkable taylorinteg(
+        lorenz!, $q0, $t0, $tf2, $_order, $_abstol, #=Val(true) o Val(false),=# $pars, -1.0, maxsteps=$maxsteps1)
+SUITE["Lorenzlyap"]["Lorenzlyap1-3"] = @benchmarkable taylorinteg(
+        lorenz!, $q0, $t0, $tf3, $_order, $_abstol, #=Val(true) o Val(false),=# $pars, -1.0, maxsteps=$maxsteps1)
```
lbenet, line 34: Please delete the comments, i.e. `#=Val(true) o Val(false),=#`, as well as the `-1.0,`... none is needed.

Also, since you are trying to benchmark the computation of the Lyapunov exponents, please use `lyap_taylorinteg` instead of `taylorinteg`.

lbenet (changes requested):
Thansk @AdrianMI49 for this addition to the benchmarks, and sorry for the delay to review it. 

I have left few comments, most are related to style, so they are not so important. There are others that do matter since they allow to overcome the errors that arise. In particular, note that your aim is to test `lyap_taylorinteg` rather than `taylorinteg`.

lbenet (approved):

lbenet:
Thanks a lot @AdrianMI49. Merging!

Issue #160, opened 2023-04-17
Module loading sequence hang with OrdinaryDiffEq.jl


vram0gh:
When using `OrdinaryDiffEq.jl` and `TaylorIntegration.jl` together, the sequence of using statements always has to be:

```julia
using OrdinaryDiffEq
using TaylorIntegration
```

in that order, otherwise a hang occurs. For example, this hangs:

```julia
using TaylorIntegration
using OrdinaryDiffEq
```

and so does this even though the using statements occur in sister modules:

```julia
module ModTIOnly
  using TaylorIntegration
end

module ModODEjlOnly
  using OrdinaryDiffEq
end
```

I originally opened this issue in the OrdinaryDiffEq.jl repo, https://github.com/SciML/OrdinaryDiffEq.jl/issues/1931 using the logic that since the hang occurs when loading `OrdinaryDiffeq` , but not the other way around, the proximate cause was in `OrdinaryDiffEq` and not `TaylorIntegration`, but it was closed there indicating the issue should be filed with this package.

Versions: Julia 1.8.5 on Linux/x86-64, OrdinaryDiffEq v6.50.0 , TaylorIntegration v0.11.1

lbenet:
Thanks for reporting here this problem. I can indeed reproduce this.

I'm not sure why it happens. Could it be that we use `Requires.jl` and check if DiffEqBase is loaded, but not if OrdinaryDiffEq is? Should we add both as direct dependencies? Any clues?

cc @ChrisRackauckas

ChrisRackauckas:
If you're using anything from OrdinaryDiffEq.jl then you need to `@requires` it. Though I'd highly recommend changing from Requires to the newer package extensions.

lbenet:
> If you're using anything from OrdinaryDiffEq.jl then you need to @requires it. Though I'd highly recommend changing from Requires to the newer package extensions.

That would only work for Julia >= v1.9, am I right? What about v1.8?

ChrisRackauckas:
Some issues with Requires are just not solvable, so I wouldn't worry too much.

lbenet:
@vram0gh This is fixed by #161, though the fix does not exploit the new pkg extensions...

PR #161, opened 2023-04-18
Fix require to use solve via OrdinaryDiffEq

lbenet:
This PR is aimed to solve #160 ~~following [this comment](https://github.com/PerezHz/TaylorIntegration.jl/issues/160#issuecomment-1512314166). The relevant docs are [here](https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)).~~

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TaylorIntegration"
 uuid = "92b13dbe-c966-51a2-8445-caca9f8a7d42"
 repo = "https://github.com/PerezHz/TaylorIntegration.jl.git"
-version = "0.11.1"
+version = "0.11.2"
 
 [deps]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
--- a/src/TaylorIntegration.jl
+++ b/src/TaylorIntegration.jl
@@ -28,6 +28,10 @@ function __init__()
         end
     end
 
+    @require DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e" begin
+        using OrdinaryDiffEq
+    end
+
 end
 
 end #module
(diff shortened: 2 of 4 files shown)


lbenet:
@vram0gh Despite what I wrote before, this PR does not solve #160. So I will close this attempt. The following is what I have found out.

I have been working from the REPL and exploiting the function `loadedModules` (slightly changed) from [here](https://stackoverflow.com/questions/25575406/list-of-loaded-imported-packages-in-julia), which checks what modules are being loaded in a module (`Main` and `TaylorIntegration`). I'm using the current version in master (v0.11.1).

- Case 1, `using DiffEqBase` followed by `using TaylorIntegration`: `@requires` does what it is supposed to do, i.e., includes `src/common.jl`, which loads both `DiffEqBase` (redundantly) and `OrdinaryDiffEq`. In this case, `DiffEqBase` appears loaded both in `Main` and `Taylorintegration`, and `OrdinaryDiffEq` does not appear there.

- Case 2, `using TaylorIntegration` followed by `using DiffEqBase`: Same as the previous case.

- Case 3, `using OrdinaryDiffEq` followed by `using TaylorIntegration`: `@requires` is triggered (!), since `DiffEqBase` appears among the loaded modules in `TaylorIntegration`, and `OrdinaryDiffEq` appears among those loaded in `Main`. Loading `OrdinaryDiffEq` does a `@reexport using DiffEqBase`, which seems to work.

- Case 4 (the problematic one), `using TaylorIntegration` followed by `using OrdinaryDiffEq`: Here, again, `@requires` is triggered, but hangs indefinitely as reported; interrupting with ctrl-C show the warning "Error requiring `DiffEqBase` from `TaylorIntegration`". Somehow, `using OrdinaryDiffEq` activates the `@requires` as before, but the `@reexport using DiffEqBase` is not recognized.

I think we should documented this, and leave open #160. Hopefully using the new Pkg capabilities allows to solve this...

@ChrisRackauckas @PerezHz Other ideas are welcome...

PerezHz:
Thank you @lbenet for taking a look into this! I tried an alternative solution to #160 which seems to be working; see #165

PR #165, opened 2023-05-13
Fix #160

### LINKED ISSUE #160: Module loading sequence hang with OrdinaryDiffEq.jl

PerezHz:
This PR is another proposal to fix #160. What this PR does is essentially to `@require` directly `OrdinaryDiffEq` instead of `DiffEqBase`. Indeed, since in the common interface `TaylorMethod` is a subtype of `OrdinaryDiffEqAdaptiveAlgorithm`, then `TaylorIntegration` should actually be used together with `OrdinaryDiffEq` instead of `DiffEqBase` when using the common interface, as is currently documented. Tests are passing, so if we all agree this is a viable solution then I'll update the docs; otherwise this should be ready to merge.

cc: @lbenet @vram0gh

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TaylorIntegration"
 uuid = "92b13dbe-c966-51a2-8445-caca9f8a7d42"
 repo = "https://github.com/PerezHz/TaylorIntegration.jl.git"
-version = "0.11.1"
+version = "0.12.0"
 
 [deps]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
--- a/Project.toml
+++ b/Project.toml
@@ -36,7 +36,6 @@ TaylorSeries = "0.14, 0.15"
 julia = "1"
 
 [extras]
-DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 Elliptic = "b305315f-e792-5b7a-8f41-49f472929428"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
@@ -47,4 +46,4 @@ TaylorSeries = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["DiffEqBase", "Elliptic", "InteractiveUtils", "LinearAlgebra", "Logging", "Pkg", "StaticArrays", "TaylorSeries", "Test"]
+test = ["Elliptic", "InteractiveUtils", "LinearAlgebra", "Logging", "Pkg", "StaticArrays", "TaylorSeries", "Test"]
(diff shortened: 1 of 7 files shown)


lbenet:
Thanks @PerezHz for spending some time on this. If I understand correctly, your solution is only to require `OrdinaryDiffEq` as a dependency, rather than `DiffEqBase`, right? In any case, I am in favor of adding the relevant docs.

Some questions:
- What happens if you load `TaylorIntegration` and then `DiffEqBase` (without any explicit load of `OrdinaryDiffEq`)? More concretely, is the requires triggered? (`DiffEqBase` loads `OrdinaryDiffEq` as a dependency.)
- What happens if you do the same, but loading first `DiffEqBase` and then `TaylorIntegration`?
- Finally, is there a way to exploit the Pkg extensions (`weakdeps`) in Julia 1.9?

PerezHz:
>If I understand correctly, your solution is only to require OrdinaryDiffEq as a dependency, rather than DiffEqBase, right?

Essentially, yes

> In any case, I am in favor of adding the relevant docs.

Will do then!

> What happens if you load TaylorIntegration and then DiffEqBase (without any explicit load of OrdinaryDiffEq)? More concretely, is the requires triggered? (DiffEqBase loads OrdinaryDiffEq as a dependency.)

Doing either `using TaylorIntegration; using DiffEqBase` or `using DiffEqBase; using TaylorIntegration` works fine from a fresh julia session with the proposed solution. We do overload `DiffEqBase.solve`, so we have to keep it as an explicit dependency.

> More concretely, is the requires triggered? (DiffEqBase loads OrdinaryDiffEq as a dependency.)

As far as I understand it's rather DiffEqBase that is loaded as a dependency of OrdinaryDiffEq, and the former is actually `@reexport`ed by the latter when the `@require` is triggered when loading OrdinaryDiffEq. The `@require` is only triggered when OrdinaryDiffEq is loaded (either before of after) TaylorIntegration. With the proposed change users who want to use TaylorIntegration via the DiffEq common interface, will now have to do either `using TaylorIntegration, OrdinaryDiffEq` or `using OrdinaryDiffEq, TaylorIntegration` (i.e., no longer need to load DiffEqBase), with the proposed solution.

> What happens if you do the same, but loading first DiffEqBase and then TaylorIntegration?

Works fine

> Finally, is there a way to exploit the Pkg extensions (weakdeps) in Julia 1.9?

Sure there is, and this should be the next step, but I would suggest merging this change, which will help users of julia 1.6 and above, and then do another update via package extensions.

lbenet:
Thanks for answering my questions. I suggest to include the changes in the docs, merge it and tag a new minor version

PR #167, opened 2023-05-20
Fix #166

### LINKED ISSUE #166: Conditionally load OrdinaryDiffEq via weakdeps
We should use weakdeps for julia 1.9+, while using requires for earlier julia versions.

PerezHz:
This PR allows using weakdeps to conditionally load OrdinaryDiffEqs (and associated "sub-weakdeps") for julia 1.9+, while retaining requires for earlier julia versions.


PerezHz:
As a workaround to the `TaylorMethod` struct currently exported by the common interface, I took inspiration from [this comment](https://discourse.julialang.org/t/question-about-new-package-extensions/92794/5) on Discourse. So as it is now, this PR is ready for review @lbenet.

ext/TaylorIntegrationDiffEq.jl
```diff
@@ -0,0 +1,313 @@
...
+
+
+
+abstract type TaylorAlgorithm <: OrdinaryDiffEqAdaptiveAlgorithm end
+struct TaylorMethodParams <: TaylorAlgorithm
+    order::Int
+    parse_eqs::Bool
+end
+
+import TaylorIntegration: TaylorMethod
```
lbenet, line 44: Nice work around! I was stuck precisely here...

lbenet:
LGTM, and tests pass, so let's go ahead and merge this PR. I suggest we bump a minor version.

PR #168, opened 2023-06-18
Fix dense output

PerezHz:
This PR fixes a subtle bug in the dense output polynomials. This will require bumping a minor version, and perhaps adding a warning to users, since this change may break downstream code.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TaylorIntegration"
 uuid = "92b13dbe-c966-51a2-8445-caca9f8a7d42"
 repo = "https://github.com/PerezHz/TaylorIntegration.jl.git"
-version = "0.13.0"
+version = "0.14.0"
 
 [deps]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
--- a/src/TaylorIntegration.jl
+++ b/src/TaylorIntegration.jl
@@ -20,6 +20,17 @@ include("rootfinding.jl")
 include("common.jl")
 
 function __init__()
+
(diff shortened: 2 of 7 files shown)


PerezHz:
@lbenet tests are not currently passing on nightly, but otherwise this is now ready for review.

lbenet (approved):

lbenet:
Thanks a lot! LGTM! While I agree to merge this, I suggest to wait a little bit to (have some time and) check some packages that depend on TaylorIntegration, and in case of necessity, also update them.

lbenet:
Note that there seems to be [a test](https://github.com/PerezHz/TaylorIntegration.jl/actions/runs/5404746318/jobs/9819339149?pr=168#step:6:410) related to `@taylorize` which is not passing

PerezHz:
> Note that there seems to be [a test](https://github.com/PerezHz/TaylorIntegration.jl/actions/runs/5404746318/jobs/9819339149?pr=168#step:6:410) related to @taylorize which is not passing

Indeed! I was thinking to maybe wait for nightlies to stabilize, or do you think it's worth to deal with it already?

lbenet:
Let me have a look on it tomorrow...

lbenet:
It seems that Julia nightly includes additional information related to the line that produced certain code, which is causing the problem.

PerezHz:
I think I found the related change in julia: https://github.com/JuliaLang/julia/pull/44995

PerezHz:
Thank you for the suggestion @lbenet! I think I found a workaround with `Base.remove_linenums!`, which in principle should work for all tested versions...

PerezHz:
Got the green lights! I'll merge and trigger the new release!

PR #170, opened 2023-06-30
Add root-finding with dense output `taylorinteg` methods

PerezHz:


PerezHz:
@lbenet this is now ready for review. I have made some local tests and with this PR we can indeed remove the remaining `neosinteg` method. So once we merge this, I propose to update https://github.com/PerezHz/NEOs.jl/pull/25, and make progress towards merging it.

PerezHz:
Since tests are passing, I will go ahead and merge this just so that we can try to move forward with https://github.com/PerezHz/NEOs.jl/pull/25; we can revisit this if necessary.

PR #171, opened 2023-06-30
WIP: Optimize `@taylorize` for jet transport integrations

lbenet:
Currently, `@taylorize` optimizes allocations with respect to *time* (independent variable), but not with respect to the jet-transport variables (e.g., *initial conditions* or parameters), when such integration is performed. This makes that the number of allocations grow and performance is bad. 

This draft aims at improving that. My current idea is to create two new parsed functions which are specialized for `Taylor1{TaylorN{S}}` entries, mimicking what we do now. As I see it now, a new inner for-loop is needed (for each fixed order in the *time* loop), which runs over the order of the jet-transport variables.


lbenet:
Closing, in favor of improvements in TS

PR #173, opened 2023-11-12
Small fix in docs

PerezHz:
After manually releasing TaylorIntegration v0.14.3 due to a previous TagBot error, docs were not building properly, due to a small error in a `@ref` to `RetAlloc`. I went ahead and also updated the discussion on the PCR3BP integrations where the performance of TaylorIntegration is compared with `Vern9()`.

### DIFF
--- a/docs/src/common.md
+++ b/docs/src/common.md
@@ -244,9 +244,9 @@ xlabel!("t")
 ylabel!("dE")
 ```
 We notice that the Jacobi constant absolute error for the `TaylorMethod(25)`
-solution remains bounded below ``10^{-13}``. The `Vern9()` solution is, at the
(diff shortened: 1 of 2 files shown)


lbenet:
Thanks @PerezHz for noticing the problem with the docs, and for addressing it. (Also, thanks for tagging 0.14.3!) 

I notice that there are many tests that fail, in all platforms, and also that we are not testing Julia 1.10. I think it is a good opportunity to solve those problematic tests here, and test with Julia 1.10-rc1, in view that the rnew Julia elease will happen soon; alternatively, openning an issue is good enough for now.

lbenet:
20 seconds too late 😄

PR #178, opened 2024-01-23
Automatic Domain Splitting

LuEdRaMo:
At present, `TaylorIntegration` (`TI`) supports jet transport (JT) "naively". This means that we assume that polynomials converge both with respect to time and JT variables. However, we don't enforce it. The time step is a control mechanism in the time domain, but there is no analog in the JT space. As a result, there are examples where the coefficients grow too large, causing the polynomial to no longer represent the solution accurately.

To solve this issue, this PR introduces a new `taylorinteg` method that implements a technique called **Automatic Domain Splitting** (ADS), proposed by [Wittig, et. al. (2015)](https://doi.org/10.1007/s10569-015-9618-3). The idea is simple: we begin with a domain where we can trust JT works (`ADSDomain`). As we propagate, whenever an error metric exceeds a tolerance (`stol`), the current polynomial splits into two polynomials, each representing half of the original domain. The new `taylorinteg` method returns a customized binary tree data structure (`ADSBinaryNode`).

For instance, [Wittig, et. al. (2015)](https://doi.org/10.1007/s10569-015-9618-3) studied a 2D Kepler problem. In the animation below, the left panel depicts a Monte Carlo propagation of a small box of initial conditions. On the other hand, the middle panel shows the "naive" JT propagation that we would perform in the current TI version. Finally, the right panel shows the equivalent ADS integration. It is noticeable that the naive JT fails to represent the solution accurately after the second revolution.

https://github.com/PerezHz/TaylorIntegration.jl/assets/73906617/cee00546-91dd-41ff-9d1f-faa698b9795a


lbenet:
Thanks a lot @LuEdRaMo for this initiative! Please let us know when shall we review your PR. Incidentally, the video was not uploaded properly.

LuEdRaMo:
> Thanks a lot @LuEdRaMo for this initiative! Please let us know when shall we review your PR. Incidentally, the video was not uploaded properly.

I've changed the video format, it should play properly now.

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,327 @@
...
+
+    # Integration
+    while sign_tstep*t0 < sign_tstep*tmax
+
+        for (k, node) in enumerate(Leaves(nv))
+            @inbounds for i in 1:M
+                @inbounds x[i, k] = Taylor1( node.x[i], order )
+                @inbounds dx[i, k] = Taylor1( zero(node.x[i]), order )
+            end
+            xaux = Vector{Taylor1{TaylorN{T}}}(undef, M)
```
lbenet, line 244: I think it is better to define `xaux` outside the loop (perhaps as a matrix), and reuse it.

test/ads.jl
lbenet, line 1: I think this test should be incorporated in the documentation too, with some very basic explanation.

lbenet (commented):
Thanks a lot @LuEdRaMo, very nice work! I've left some comments, mostly questions, and perhaps a couple of suggestions, among other, adding a section on the documentation, and perhaps for improving marginally the performance.

I see this PR is progressing nicely, and while it is not necesary for merging this PR, I still pose the question: are you planing adding in this PR the possibility of using common interface to run it from DifferentialEquations? (If not, perhaps it is a good idea to open an issue about this for the future.)

src/adsbinarynode.jl
```diff
@@ -0,0 +1,381 @@
...
+    @assert 1 <= i <= N
+    mid = (s.lo[i] + s.hi[i])/2
+    a = ADSDomain{N, T}(
+        s.lo,
+        SVector{N, T}(i == j ? mid : s.hi[j] for j in 1:N)
+    )
+    b = ADSDomain{N, T}(
+        SVector{N, T}(i == j ? mid : s.lo[j] for j in 1:N),
+        s.hi
+    )
```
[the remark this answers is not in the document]
LuEdRaMo: I don't see a way to avoid the two `fors`, given that `SVector` is immutable. However, this is not that bad, since the `SVector` is built upon a generator and this function is not critical to the performance.

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,327 @@
...
+        for j in eachindex(idxs)
+            for k in eachindex(idxs[j])
+                ys[k, idxs[j][k]+1] += abs(P.coeffs[i].coeffs[j])
+            end
+        end
+    end
+    # Initial parameters
+    p0 = ones(T, 2)
+    # Orders
+    xs = collect(0:varorder)
```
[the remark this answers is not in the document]
LuEdRaMo: You're right, `xs` can stay as a range. I'll change that in the next commits.

src/adsbinarynode.jl
```diff
@@ -0,0 +1,381 @@
...
+    push!(ts, n.t)
+    timesvector!(n.left, ts)
+    timesvector!(n.right, ts)
+    nothing
+end
+
+timesvector!(n::Nothing, ts::Set{T}) where {T <: Real} = nothing
+
+# Auxiliary evaluation methods
+function _eval(p::SVector{M, Taylor1{TaylorN{T}}}, dt::T) where {M, T <: Real}
```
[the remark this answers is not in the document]
LuEdRaMo: `_eval` do not evaluate a node or tree, so I think a better name would be `_adseval`.

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,327 @@
...
+    nv = get_numvars()
+    # Size per variable per element of P
+    norms = Matrix{T}(undef, nv, length(P))
+    for i in eachindex(P)
+        norms[:, i] .= size_per_variable(P[i])
+    end
+    # Sum over element of P
+    M = sum(norms, dims = 2)
+
+    return M
```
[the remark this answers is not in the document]
LuEdRaMo: Done.

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,327 @@
...
+    t = t0 + Taylor1( T, order )
+    x = Array{Taylor1{TaylorN{T}}}(undef, M)
+    dx = Array{Taylor1{TaylorN{T}}}(undef, M)
+    @inbounds for i in eachindex(q0)
+        @inbounds x[i] = Taylor1( q0[i], order )
+        @inbounds dx[i] = Taylor1( zero(q0[i]), order )
+    end
+
+    # Determine if specialized jetcoeffs! method exists
+    parse_eqs, rv = _determine_parsing!(parse_eqs, f!, t, x, dx, params)
```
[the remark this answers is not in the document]
LuEdRaMo: Since the integration of the splits is done sequentially, we can reuse `rv`. In case we wanted to propagate the splits in parallel, we might need a `rv` per task.

LuEdRaMo:
> Thanks a lot @LuEdRaMo, very nice work! I've left some comments, mostly questions, and perhaps a couple of suggestions, among other, adding a section on the documentation, and perhaps for improving marginally the performance.

Thanks @lbenet for your suggestions.

> I see this PR is progressing nicely, and while it is not necesary for merging this PR, I still pose the question: are you planing adding in this PR the possibility of using common interface to run it from DifferentialEquations? (If not, perhaps it is a good idea to open an issue about this for the future.)

I'm not familiar with the `DifferentialEquations` interface, so I think we should open an issue and leave that for a future PR.

src/adstaylorinteg.jl
LuEdRaMo: Done.

src/adsbinarynode.jl
```diff
@@ -0,0 +1,400 @@
...
+# Outer constructors
+ADSDomain(lo::SVector{N, T}, hi::SVector{N, T}) where {N, T <: Real} = ADSDomain{N, T}(lo, hi)
+
+function ADSDomain(x::Tuple{T, T}...) where {T <: Real}
+    N = length(x)
+    return ADSDomain{N, T}(
+        SVector{N, T}(x[i][1] for i in eachindex(x)),
+        SVector{N, T}(x[i][2] for i in eachindex(x))
+    )
+end
```
lbenet, line 44: Testing locally, I obtain that this method is not passing an Aqua test, more specifically, 
```julia
ua = Aqua.detect_unbound_args_recursively(TaylorIntegration)
```
signals this method. (`ua` should have length 0.)

Project.toml
```diff
@@ -4,10 +4,12 @@ repo = "https://github.com/PerezHz/TaylorIntegration.jl.git"
 version = "0.15.1"
 
 [deps]
+AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 Espresso = "6912e4f1-e036-58b0-9138-08d1e6358ea9"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+LsqFit = "2fda8390-95c7-5789-9bda-21331edee243"
```
lbenet, line 12: The new added packages require a `compat` entry too.

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,327 @@
+# This file is part of the TaylorIntegration.jl package; MIT licensed
+
+using TaylorSeries: generate_index_vectors
```
lbenet, line 3: I suggest to remove this line; my point is that the actual number of variables and order have been set, so this information has already been calculated and is stored in memory.
```suggestion
```

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,327 @@
...
+function size_per_variable(P::TaylorN{T}) where {T <: Real}
+    # Jet transport order
+    varorder = P.order
+    # Number of variables
+    nv = get_numvars()
+    # Absolute sum per variable per order
+    ys = zeros(T, nv, varorder+1)
+
+    for i in eachindex(P.coeffs)
+        idxs = generate_index_vectors(nv, i-1)
```
lbenet, line 37: Related to the change suggested above, you should then change this line to:
```suggestion
        idx = TaylorSeries.coeff_table[i]
```

src/adsbinarynode.jl
LuEdRaMo: Following Aqua's documentation, I added a trivial constructor `ADSDomain()`.

Project.toml
LuEdRaMo: Done.

src/adstaylorinteg.jl
LuEdRaMo: Thanks for the suggestion, I didn't know about `coeff_table`.

lbenet:
@PerezHz Could we reintroduce the ci tests when pushing? Currently, only the docs are build... (This is not necessary for this PR, but certainly useful.)

PerezHz:
Fix is now in place in main branch.

PerezHz:
Took the liberty to update this branch, current expectation is avoid tests failing due to the Aqua issue mentioned in #189.

PerezHz:
Modulo some test failures on nightly (might be worth to re-run and/or report afterwards) tests are passing now 👍

src/adsbinarynode.jl
```diff
@@ -0,0 +1,402 @@
...
+    hi::SVector{N, T}
+    # Inner constructor
+    function ADSDomain{N, T}(lo::SVector{N, T}, hi::SVector{N, T}) where {N, T <: Real}
+        @assert all(hi .> lo) "Each upper bound must be greater than the respective lower bound"
+        new{N, T}(lo, hi)
+    end
+end
+
+# Outer constructors
+ADSDomain(lo::SVector{N, T}, hi::SVector{N, T}) where {N, T <: Real} = ADSDomain{N, T}(lo, hi)
```
PerezHz, line 36: I'd suggest to add a couple of unit tests for the constructors in lines 36 and 38

src/adsbinarynode.jl
```diff
@@ -0,0 +1,402 @@
...
+    for i in eachindex(x)
+        x[i] = string(SVector{2, T}(s.lo[i], s.hi[i]))
+    end
+    print(io, join(x, "×"))
+end
+
+# Return the diameter of each direction in s
+diams(s::ADSDomain{N, T}) where {N, T <: Real} = s.hi - s.lo
+# Return the lower bounds in s
+infimums(s::ADSDomain{N, T}) where {N, T <: Real} = s.lo
```
PerezHz, line 62: Since `infimums` is essentially a getter for the `lo` field, I'd suggest to either rename this method to `lo`, or rename the field to `infima`. Either way, I think it's more common to see `infima` for the plural (same comment for `supremums`).

src/adsbinarynode.jl
```diff
@@ -0,0 +1,402 @@
...
+end
+
+_countnodes(n::Nothing, t::T) where {T <: Real} = 0
+
+"""
+    timeshift!(n::ADSBinaryNode{N, M, T}, dt::T) where {N, M, T <: Real}
+
+Shift the time of every node in `n` by `dt`.
+"""
+function timeshift!(n::ADSBinaryNode{N, M, T}, dt::T) where {N, M, T <: Real}
```
PerezHz, line 237: I could not find tests covering `timeshift!` or else being used throught the code, so I'd suggest to either remove this function, or it's deemed useful, to add tests for it.

test/ads.jl
```diff
@@ -0,0 +1,129 @@
+using TaylorIntegration
+using Test
+
+@testset "Automatic Domain Splitting" begin
```
PerezHz, line 4: I'd suggest to add a test where we know there will be no splitting (a linear harmonic oscillator for example) and see that we don't produce splitting there. Also, a test involving the RTBP for unbounded orbits (aka CR3BP) can be interesting.

PerezHz (commented):
Sorry for taking long to get to this, it's an amazing contribution @LuEdRaMo! I left a couple of comments and will continue taking a look in the next few days.

test/ads.jl
```diff
@@ -0,0 +1,129 @@
...
+    # Splitting tolerance
+    local stol = 0.008
+    # Absolute tolerance
+    local abstol = 1e-20
+    # Dynamical function parameters
+    local params = nothing
+    # Maximum allowed splits
+    local maxsplits = 25
+    # Maximum allowed steps
+    local maxsteps = 1_000
```
PerezHz, line 43: Since `@testset begin ... end` introduces a scope, I think it's safe to remove the `local`s here.

src/adsbinarynode.jl
```diff
@@ -0,0 +1,402 @@
...
+    parent::ADSBinaryNode{N, M, T}, s::ADSDomain{N, T}, t::T,
+    x::SVector{M, TaylorN{T}}, p::SVector{M, Taylor1{TaylorN{T}}}
+    ) where {N, M, T <: Real}
+    isnothing(parent.right) || error("Right child is already assigned")
+    node = ADSBinaryNode{N, M, T}(s, t, x, p, parent.depth + 1, parent)
+    parent.right = node
+end
+
+# AbstractTrees interface
+function children(node::ADSBinaryNode{N, M, T}) where {N, M, T <: Real}
```
PerezHz, line 163: For all the method overloads from `Base` and `AbstractTrees` I'd suggest to switch to explicitly prefixing the import so that it's clear which methods we are overloading from where. E.g.,

```julia
import Base, AbstractTrees

function AbstractTrees.children(...)
    # (overload for ADSBinary Node)
end

function AbstractTrees.printnode(...)
    # (overload for ADSBinaryNode)
end

function Base.IteratorEltype(...)
    # overload for ADSBinaryNode
end

# etc...
```

PerezHz (commented):
(Added a few more comments...)

PerezHz:
Question: it can be interesting to do some root-finding on ADS integrations, is that support already added, or is that still in the works?

PerezHz:
> I see this PR is progressing nicely, and while it is not necesary for merging this PR, I still pose the question: are you planing adding in this PR the possibility of using common interface to run it from DifferentialEquations? (If not, perhaps it is a good idea to open an issue about this for the future.)

Regarding this point, I'd agree to leave this open for now (I'm starting to do a bit of a re-design of the TaylorIntegration internals which might help us add this support more easily).

src/adsbinarynode.jl
```diff
@@ -0,0 +1,402 @@
...
+using StaticArrays: SVector
+import AbstractTrees: HasNodeType, NodeType, ParentLinks, StoredParents, TreeIterator,
+                      children, nodetype, nodevalue, parent, printnode, PreOrderDFS
+import Base: IteratorEltype, HasEltype, in, eltype, split
+
+if !isdefined(Base, :isnothing)        # Julia 1.0 support
+    using AbstractTrees: isnothing
+end
+
+export ADSDomain
```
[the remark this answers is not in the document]
LuEdRaMo: Done.

test/ads.jl
LuEdRaMo: Done.

test/ads.jl
```diff
@@ -0,0 +1,129 @@
...
+
+@testset "Automatic Domain Splitting" begin
+    using TaylorIntegration: ADSBinaryNode, countnodes, timesvector
+    using StaticArrays: SVector
+
+    # This example is based upon section 3 of
+    # https://doi.org/10.1007/s10569-015-9618-3.
+
+    # Dynamical function
+    @taylorize function kepler_eqs!(dq, q, params, t)
```
PerezHz, line 12: I suggest adding the following check in the Kepler problem: Say we take an initial (non-nominal) condition `q1` in the initial domain defined by `q0` and `dom`, which will have initial Keplerian elements say `a1`, `e1` (semimajor axis and eccentricity, etc.). Then, at the end of the ADS integration, the propagated initial condition `q2` will have a semimajor axis `a2`, eccentricity `e2`, etc. And we should have that `a1` and `a2` (resp., `e1` and `e2`) are equal within the order of magnitude of the splitting tolerance `stol`.

test/ads.jl
```diff
@@ -0,0 +1,129 @@
...
+        dq[1] = q[3]
```
PerezHz, line 13: Another interesting check/question can be the following one: write down an initial condition `q1` as above in terms of Cartesian position and velocity, while defining the JT variables in terms of perturbations in the Keplerian elements `δa`, `δe`, etc. Then, upon propagating this initial condition on a domain `dom`, can we expect almost no splitting, since in this setup the elements are preserved?

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,326 @@
...
+    p1 = SVector{M, Taylor1{TaylorN{T}}}(
+        Taylor1( map(z -> z(v_1), p[i].coeffs), order ) for i in eachindex(p)
+    )
+    # Right half
+    x2 = SVector{M, TaylorN{T}}(x[i](v_2) for i in eachindex(x))
+    p2 = SVector{M, Taylor1{TaylorN{T}}}(
+        Taylor1( map(z -> z(v_2), p[i].coeffs), order ) for i in eachindex(p)
+    )
+
+    return s1, x1, p1, s2, x2, p2
```
PerezHz, line 149: If we passed `dt` as an argument to `split`, then here we could return a tuple of `ADSBinaryNode`s, i.e.,
```julia
    return ADSBinaryNode(s1, node.t+dt, x1, p1), ADSBinaryNode(s2, node.t+dt, x2, p2)
```
and then set those below in `split!` in the corresponding calls to `leftchild!` and `rightchild!`; though I think it'd require defining new methods:

```julia
function leftchild!(parent::ADSBinaryNode, newnode::ADSBinaryNode)
    # set `newnode` as left child of `parent`
end
```

Exploiting the type parametrization in the `ADSBinaryNode` constructor can help with type-stability.

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,326 @@
...
+    # Split criteria for each element of x
+    mask = adsnorm.(x)
+    # Split
+    if nsplits < maxsplits && any(mask .> stol)
+        # Split
+        s1, x1, p1, s2, x2, p2 = split(node, x, p)
+        # Left half
+        leftchild!(node, s1, node.t + dt, x1, p1)
+        # Right half
+        rightchild!(node, s2, node.t + dt, x2, p2)
```
PerezHz, line 171: Related to my comment above, what I'm proposing is:
```suggestion
        # Split
        lchildnode, rchildnode = split(node, x, p, dt) # note the addition of `dt` to the `split` call
        # Left half
        leftchild!(node lchildnode) # call to proposed leftchild! method
        # Right half
        rightchild!(node, rchildnode) # call to proposed rightchild! method
```

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,326 @@
...
+"""
+function adsnorm(P::TaylorN{T}) where {T <: Real}
+    # Jet transport order
+    varorder = P.order
+    # Absolute sum per order
+    ys = Vector{T}(undef, varorder+1)
+
+    for i in eachindex(ys)
+        ys[i] = sum(abs, P.coeffs[i].coeffs)
+    end
```
PerezHz, line 108: Suggestion: does it make sense here to pass `ys` as an arg to `adsnorm` so that `ys` is only updated here without having to allocate it on every call?

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,326 @@
...
+"""
+    size_per_variable(P::TaylorN{T}) where {T <: Real}
+    size_per_variable(P::AbstractVector{TaylorN{T}}) where {T <: Real}
+
+Return the truncation error of each jet transport variable in `P`.
+
+!!! reference
+    See section 3 of https://doi.org/10.1007/s10569-015-9618-3.
+"""
+function size_per_variable(P::TaylorN{T}) where {T <: Real}
```
PerezHz, line 27: Given that it refers to truncation error, I'd suggest to rename `size_per_variable` to something like `trunc_error`, or `error_per_variable`?

test/ads.jl
PerezHz: Another check I think worth adding: take a number (say, 16) of initial conditions in the border of the domain (e.g., in the corners), propagate them individually without ADS, then check the difference at the end of the integration wrt ADS propagation.

src/adstaylorinteg.jl
LuEdRaMo: I'll rename this function to `truncerror`.

src/adsbinarynode.jl
```diff
@@ -0,0 +1,402 @@
...
+function show(io::IO, s::ADSDomain{N, T}) where {N, T <: Real}
+    x = Vector{String}(undef, N)
+    for i in eachindex(x)
+        x[i] = string(SVector{2, T}(s.lo[i], s.hi[i]))
+    end
+    print(io, join(x, "×"))
+end
+
+# Return the diameter of each direction in s
+diams(s::ADSDomain{N, T}) where {N, T <: Real} = s.hi - s.lo
```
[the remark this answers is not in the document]
LuEdRaMo: Done.

src/adsbinarynode.jl
LuEdRaMo: I think the second option sounds cooler, so I'll rename this functions to `infima` and `suprema` respectively.

src/adstaylorinteg.jl
LuEdRaMo: Thanks for the suggestion! Now `split` returns a tuple of `ADSBinaryNode`.

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,326 @@
...
+    # No split
+    else
+        leftchild!(node, node.s, node.t + dt, x, p)
+    end
+
+    return nsplits
+end
+
+function taylorinteg(
+    f!, q0::SVector{M, TaylorN{T}}, s::ADSDomain{N, T}, t0::T, tmax::T,
```
PerezHz, line 183: I'm thinking it can make sense to add the reference initial condition `q0` as a field to `ADSDomain`? Then, an ADS integration can be a call to `taylorinteg` on an initial condition `<: ADSDomain`, in a similar way as JT is a dispatch on `Taylor1{TaylorN{T}}`.

src/adsbinarynode.jl
LuEdRaMo: Done.

src/adsbinarynode.jl
LuEdRaMo: `timeshift!` will be useful in the future to set the epoch of an ADS integration in NEOs. Meanwhile, I've added a couple of tests for this function.

src/adsbinarynode.jl
```diff
@@ -0,0 +1,402 @@
...
+
+ParentLinks(::Type{<:ADSBinaryNode{N, M, T}}) where {N, M, T <: Real} = StoredParents()
+
+parent(n::ADSBinaryNode{N, M, T}) where {N, M, T <: Real} = n.parent
+
+NodeType(::Type{<:ADSBinaryNode{N, M, T}}) where {N, M, T <: Real} = HasNodeType()
+nodetype(::Type{<:ADSBinaryNode{N, M, T}}) where {N, M, T <: Real} = ADSBinaryNode{N, M, T}
+
+# For TreeIterator
+IteratorEltype(::Type{<:TreeIterator{ADSBinaryNode{N, M, T}}}) where {N, M, T <: Real} = HasEltype()
```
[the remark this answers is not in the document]
LuEdRaMo: I agree that this may be redundant, but we should leave it as it is following this [note](https://juliacollections.github.io/AbstractTrees.jl/stable/iteration/#Interface) from `AbstractTrees`.

src/adstaylorinteg.jl
lbenet: Naive question: is there an issue with type stability?

src/adsbinarynode.jl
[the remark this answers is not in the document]
PerezHz: I agree we should keep it; I would only suggest to add a comment in the code explaining this together with the link you posted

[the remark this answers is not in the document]
LuEdRaMo: Done.

src/adsbinarynode.jl
PerezHz: Regarding this point (unless constructing an `ADSDomain` without arguments has a specific use), I would be in favor of `ADSDomain()` returning an error? Simply because throughout the code the constructor with arguments is the one that will be used to construct `ADSDomain` variables, and if by any chance we call the wrong constructor we might end up with silent errors / hard-to-catch bugs.

src/adstaylorinteg.jl
PerezHz: Or rather, I think we should try to figure out what type (struct) (let's call it `X`) we can give as an initial condition, such that at every new time step `t`, the solution at time `t` is a new instance of type `X`. Maybe it makes more sense then to give an `ADSBinaryNode` root node as an initial condition (in other words, would `X=ADSBinaryNode` be a good choice here)? Then we could return the solution as a `Vector` of time steps and the solution at each `t` as a `Vector{X}`.

test/ads.jl
lbenet: A possibility is to check that the first few steps run without any splitting

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,327 @@
...
+    # Absolute sum per variable per order
+    ys = zeros(T, nv, varorder+1)
+
+    for i in eachindex(P.coeffs)
+        idxs = TaylorSeries.coeff_table[i]
+        for j in eachindex(idxs)
+            for k in eachindex(idxs[j])
+                ys[k, idxs[j][k]+1] += abs(P.coeffs[i].coeffs[j])
+            end
+        end
```
lbenet, line 41:
```suggestion
        for j in eachindex(idxs)
            coef = abs(P.coeffs[i].coeffs[j])
            for k in eachindex(idxs[j])
                ys[k, idxs[j][k]+1] += coef
            end
        end
```

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,327 @@
...
+function truncerror(P::AbstractVector{TaylorN{T}}) where {T <: Real}
+    # Number of variables
+    nv = get_numvars()
+    # Size per variable per element of P
+    norms = Matrix{T}(undef, nv, length(P))
+    for i in eachindex(P)
+        norms[:, i] .= truncerror(P[i])
+    end
+    # Sum over element of P
+    return sum(norms, dims = 2)
```
lbenet, line 71:
```suggestion
    norms = Vector{T}(undef, nv)
    # Sum over elements of P
    norms .= sum(truncerror.(P[:]))
    return norms
```

src/adstaylorinteg.jl
```diff
@@ -0,0 +1,327 @@
...
+"""
+function adsnorm(P::TaylorN{T}) where {T <: Real}
+    # Jet transport order
+    varorder = P.order
+    # Absolute sum per order
+    ys = Vector{T}(undef, varorder+1)
+
+    for i in eachindex(ys)
+        ys[i] = sum(abs, P.coeffs[i].coeffs)
+    end
```
lbenet, line 108:
```suggestion
    ys = norm.((P[i] for i in eachindex(P)), Ref(1))
```

src/adstaylorinteg.jl
LuEdRaMo: Done.

src/adstaylorinteg.jl
LuEdRaMo: Done.

src/adstaylorinteg.jl
LuEdRaMo: Done. Thank you for the suggestions.

lbenet:
#192 has been merged (and the process of registering the new version is on the way), so I think you should rebase and try to adapt this PR to those changes.

LuEdRaMo:
@lbenet @PerezHz I've adapted this PR to current master. There are still some things pending, like adding more tests and documenting the ADS integrator, but before that I'd like to hear if you have any comments.

lbenet:
Thanks a lot @LuEdRaMo; I'll try to work on this this afternoon...

lbenet (commented):
Thanks a lot @LuEdRaMo, this is truly a very nice new addition! 

I have left some comments, mostly related to improving allocations, which in my opinion are partially related to using some recursive functions (aside the exponential growing number of the computed regions). There are also few questions, which are purely related to my ignorance on using AbstractTrees.

Also, as you pointed out, test coverage should be improved.

src/TaylorIntegration.jl
```diff
@@ -7,11 +7,17 @@ using Reexport
 using LinearAlgebra
 using Markdown
 using InteractiveUtils #: methodswith
+using StaticArrays: SVector
+using AbstractTrees: StoredParents, HasNodeType, TreeIterator, PreOrderDFS, Leaves,
+      getroot
+using LsqFit: curve_fit
+import AbstractTrees
```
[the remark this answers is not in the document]
LuEdRaMo: We don't, `using AbstractTrees` is enough.

src/ads.jl
```diff
@@ -0,0 +1,574 @@
...
+countnodes(::Nothing, arg) = 0
+
+function countnodes(n::ADSTaylorSolution{T, N, M}, k::Int) where {T <: Real, N, M}
+    @assert k >= 0
+    iszero(k) && return 1
+    return countnodes(n.left, k-1) + countnodes(n.right, k-1)
+end
+
+function countnodes(n::ADSTaylorSolution{T, N, M}, t::T) where {T <: Real, N, M}
+    isnothing(n.left) && return Int(t == n.t)
```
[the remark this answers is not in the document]
LuEdRaMo: No, I arbitrarily decided that `n.left` is used for the normal time evolution while `n.right` is only used when there is a split. In other words, all nodes (except leaves) have `n.left` different from nothing, while only split nodes have a non trivial `n.right`.

src/ads.jl
```diff
@@ -0,0 +1,574 @@
...
+    nothing
+end
+
+# Function-like callability method
+(n::ADSTaylorSolution{T, N, M})(t::U,
+    s::AbstractVector{T}) where {T <: Real, U <: Number, N, M} = evaltree(n, t, s)
+
+## ADS integrator
+
+# Exponential model y(t) = A * exp(B * t) used to estimate the truncation error
```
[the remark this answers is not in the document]
LuEdRaMo: It is used to estimate the truncation error of the expansion as `y(order + 1)`.

src/ads.jl
```diff
@@ -0,0 +1,574 @@
...
+    ys = zeros(T, nv, varorder+1)
+
+    for i in eachindex(P.coeffs)
+        idxs = TaylorSeries.coeff_table[i]
+        for j in eachindex(idxs)
+            coef = abs(P.coeffs[i].coeffs[j])
+            for k in eachindex(idxs[j])
+                ys[k, idxs[j][k]+1] += coef
+            end
+        end
```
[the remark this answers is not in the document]
LuEdRaMo: Done.

src/ads.jl
```diff
@@ -0,0 +1,574 @@
...
+"""
+    adsnorm(P::TaylorN{T}) where {T <: Real}
+
+Return the truncation error of `P`.
+"""
+function adsnorm(P::TaylorN{T}) where {T <: Real}
+    # Jet transport order
+    varorder = P.order
+    # Absolute sum per order
+    ys = norm.((P[i] for i in eachindex(P)), Ref(1))
```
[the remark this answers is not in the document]
LuEdRaMo: Done.

src/ads.jl
```diff
@@ -0,0 +1,574 @@
...
+    # Shift expansion point
+    r = getroot(node)
+    v_1[j] = v_1[j]/2 + r.lo[j]/2
+    v_2[j] = v_2[j]/2 + r.hi[j]/2
+    # Left half
+    x1 = SVector{M, TaylorN{T}}(x0[i](v_1) for i in eachindex(x0))
+    leftchild!(node, node.t + dt, lo1, hi1, x1)
+    # Right half
+    x2 = SVector{M, TaylorN{T}}(x0[i](v_2) for i in eachindex(x0))
+    rightchild!(node, node.t + dt, lo2, hi2, x2)
```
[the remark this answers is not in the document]
LuEdRaMo: Done.

src/ads.jl
```diff
@@ -0,0 +1,574 @@
...
+# AbstractTrees interface
+function AbstractTrees.children(node::ADSTaylorSolution)
+    if isnothing(node.left) && isnothing(node.right)
+        ()
+    elseif isnothing(node.left) && !isnothing(node.right)
+        (node.right,)
+    elseif !isnothing(node.left) && isnothing(node.right)
+        (node.left,)
+    else
+        (node.left, node.right)
```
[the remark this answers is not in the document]
LuEdRaMo: This function is used to iterate a tree, in this case `ADSTaylorSolution`, so we cannot return a `::Nothing` since `nothing` cannot be converted to `ADSTaylorSolution`.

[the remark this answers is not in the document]
lbenet: But then we do have some type unstability. My point is to somehow avoid it.

src/ads.jl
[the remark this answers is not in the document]
lbenet: Can you somewhere document this design? (Perhaps at the begining of the file.)

src/ads.jl
```diff
@@ -0,0 +1,574 @@
...
+    countnodes(n::ADSTaylorSolution{T, N, M}, t::T) where {T <: Real, N, M}
+
+Return the number of nodes at depth `k` (time `t`) starting from root node `n`.
+"""
+countnodes(::Nothing, arg) = 0
+
+function countnodes(n::ADSTaylorSolution{T, N, M}, k::Int) where {T <: Real, N, M}
+    @assert k >= 0
+    iszero(k) && return 1
+    return countnodes(n.left, k-1) + countnodes(n.right, k-1)
```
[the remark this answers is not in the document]
LuEdRaMo: I've managed to rewrite this as a `for` loop, thanks for the suggestion.

src/ads.jl
```diff
@@ -0,0 +1,574 @@
...
+end
+
+function countnodes(n::ADSTaylorSolution{T, N, M}, t::T) where {T <: Real, N, M}
+    isnothing(n.left) && return Int(t == n.t)
+    t0, tf = minmax(n.t, n.left.t)
+    t0 <= t < tf && return 1
+    return countnodes(n.left, t) + countnodes(n.right, t)
```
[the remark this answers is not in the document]
LuEdRaMo: Same as a above, this function is no longer recursive.

src/ads.jl
```diff
@@ -0,0 +1,574 @@
...
+        evaluate!(node.p, dt, view(x0, :, i))
+    end
+
+    return lo, hi, x0
+end
+
+evaltree!(::Nothing, args...) = nothing
+
+function evaltree!(n::ADSTaylorSolution{T, N, M}, t::U,
+        ns::Set{ADSTaylorSolution{T, N, M}}) where {T <: Real, U <: Number, N, M}
```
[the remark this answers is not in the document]
LuEdRaMo: I deprecated this function when rewriting this as a `for` loop.

src/ads.jl
```diff
@@ -0,0 +1,574 @@
...
+    if isnothing(n.left) && (t == n.t)
+        push!(ns, n.parent)
+    # Not last node
+    else
+        t0, tf = minmax(n.t, n.left.t)
+        if t0 <= t < tf
+            push!(ns, n)
+        else
+            evaltree!(n.left, t, ns)
+            evaltree!(n.right, t, ns)
```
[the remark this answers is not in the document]
LuEdRaMo: Same as above, this function is no longer recursive.

src/ads.jl
```diff
@@ -0,0 +1,574 @@
...
+        z = Vector{T}(ms .* s .+ ks)
+        # Eval p at transformed point
+        evaluate!(p[:, i], z, view(x0, :, i))
+    end
+    # Average
+    return sum(x0, dims = 2)[:] ./ L
+end
+
+function evaltree!(n::ADSTaylorSolution{T, N, M}, t::U, s::AbstractVector{T},
+        ns::Set{ADSTaylorSolution{T, N, M}}) where {T <: Real, U <: Number, N, M}
```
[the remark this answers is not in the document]
LuEdRaMo: Same as above, this function was deprecated in favour of rewriting `evaltree` as a `for` loop.

lbenet:
Did you check/notice any improvements (after using for loops instead of recursive functions) in the allocations or time?

src/ads.jl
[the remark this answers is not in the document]
PerezHz: Just thought that it's perhaps worth mentioning that the current proposal is following both the AbstractTrees.jl [interface](https://juliacollections.github.io/AbstractTrees.jl/stable/#The-Abstract-Tree-Interface) and AFAIK the implementation here follows closely the binary tree example provided in AbstractTrees.jl itself (see e.g. [this method](https://github.com/JuliaCollections/AbstractTrees.jl/blob/af3fabd36631efe361bc8e1ba45f9b84e898b459/test/examples/binarytree.jl#L31). This is only to say that as it is now `children` is in principle as type unstable as the example provided by them in their documentation. Maybe the compiler is somehow able to resolve this safely? @LuEdRaMo what is the output of `@code_warntype children(...)` called on, say, a leaf node and a non-leaf node?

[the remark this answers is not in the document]
LuEdRaMo: The output of `@code_warntype` is the same for all nodes:
```julia
Arguments
  #self#::Core.Const(AbstractTrees.children)
  node::ADSTaylorSolution{Float64, 2, 4}
Body::Tuple{Vararg{Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}}}
1 ── %1  = Base.getproperty(node, :left)::Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}
│    %2  = TaylorIntegration.isnothing(%1)::Bool
└───       goto #4 if not %2
2 ── %4  = Base.getproperty(node, :right)::Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}
│    %5  = TaylorIntegration.isnothing(%4)::Bool
└───       goto #4 if not %5
3 ── %7  = ()::Core.Const(())
└───       return %7
4 ┄─ %9  = Base.getproperty(node, :left)::Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}
│    %10 = TaylorIntegration.isnothing(%9)::Bool
└───       goto #7 if not %10
5 ── %12 = Base.getproperty(node, :right)::Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}
│    %13 = TaylorIntegration.isnothing(%12)::Bool
│    %14 = !%13::Bool
└───       goto #7 if not %14
6 ── %16 = Base.getproperty(node, :right)::Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}
│    %17 = Core.tuple(%16)::Tuple{Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}}
└───       return %17
7 ┄─ %19 = Base.getproperty(node, :left)::Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}
│    %20 = TaylorIntegration.isnothing(%19)::Bool
│    %21 = !%20::Bool
└───       goto #10 if not %21
8 ── %23 = Base.getproperty(node, :right)::Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}
│    %24 = TaylorIntegration.isnothing(%23)::Bool
└───       goto #10 if not %24
9 ── %26 = Base.getproperty(node, :left)::Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}
│    %27 = Core.tuple(%26)::Tuple{Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}}
└───       return %27
10 ┄ %29 = Base.getproperty(node, :left)::Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}
│    %30 = Base.getproperty(node, :right)::Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}
│    %31 = Core.tuple(%29, %30)::Tuple{Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}, Union{Nothing, ADSTaylorSolution{Float64, 2, 4}}}
└───       return %31
```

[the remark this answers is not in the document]
LuEdRaMo: AFAIK the `AbstractTrees` API does not require `children` to return a `Tuple`, so I tried returning a `Vector{ADSTaylorSolution{T, N, M}}`, it works and is type stable.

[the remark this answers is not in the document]
lbenet: If the returned tuple yields no `@code_warning` problems, I think we can go with it; otherwise, returning a `Vector` seems the way to go!

Thanks a lot for taking a look into this!

src/ads.jl
```diff
@@ -0,0 +1,554 @@
...
+        t[j] = t0 + Taylor1( T, order )
+        @inbounds for i in axes(x0, 1)
+            x0[i, j] = q0.x[i]
+            x[i, j] = Taylor1( q0.x[i], order )
+            dx[i, j] = Taylor1( zero(q0.x[i]), order )
+            xaux[i, j] = Taylor1( zero(q0.x[i]), order )
+        end
+    end
+    # IMPORTANT: each split needs its own RetAlloc
+    rvs = [deepcopy(rv) for _ in 1:maxsplits]
```
PerezHz, line 519: Is it clear why in the end a separate `RetAlloc` is needed for each split? If it's just some aliasing occurring in the arguments of `jetcoeffs!` maybe there's a way to work around it. Or maybe it's something else; either way I think it can help to understand where this is coming from.

PerezHz (commented):
Besides what @lbenet already pointed out I only left a couple of comments. Also I noticed `ADSDomain` is not there anymore and there was a comment I had left there before about a constructor but now I guess this is what they call a no-code solution 😄 Then just out of curiosity: why did you decide to get rid of `ADSDomain`? Otherwise I think this is looking pretty solid to go ahead with adding more tests and improving docs!

src/ads.jl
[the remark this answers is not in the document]
PerezHz: I think I would go for the tuple, otherwise, the performance we gain from type stability we might lose by allocating a vector each time children is called. Thanks for checking this!

src/ads.jl
```diff
@@ -0,0 +1,554 @@
...
+            mask[k] = sign_tstep * node.t < sign_tstep * tmax
+            mask[k] || continue
+            @inbounds for i in axes(x, 1)
+                x[i, k][0] = node.x[i]
+                TaylorSeries.zero!(dx[i, k], 0)
+            end
+            δt[k] = taylorstep!(Val(parse_eqs), f!, t[k], x[:, k], dx[:, k], xaux[:, k],
+                    abstol, params, rvs[k]) # δt is positive!
+            # Below, δt has the proper sign according to the direction of the integration
+            δt[k] = sign_tstep * min(δt[k], sign_tstep*(tmax-node.t))
```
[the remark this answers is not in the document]
lbenet: I think it is indeed necessary: while each time step is independent (which reflects in the `[k]` reference), the min is necessary to be at `tmax` at the end of the integration. This comes from TaylorIntegration.

src/ads.jl
lbenet: My naive guess is that you want some sort of "thread safety" in the tree context, which I'm unable to formulate properly...

src/ads.jl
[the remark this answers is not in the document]
PerezHz: Ah you're right! This changed in the last few commits, previously there was a global time step but I see it's no longer that case, so that's great!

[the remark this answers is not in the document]
PerezHz: This point, that each node carries its own time step, should probably be documented.

src/ads.jl
LuEdRaMo: My guess is that it is not a "thread safety" issue since splits are processed sequentially. When I started doing ADS integrations in `NEOs` I noticed unexpected results that were solved by introducing one `RetAlloc` per split. So, there is something in `NEOs` dynamical functions that tests here do not notice and affect the `RetAlloc` object.

LuEdRaMo:
> Then just out of curiosity: why did you decide to get rid of `ADSDomain`?

That change was motivated by one of your comments:

> I'm thinking it can make sense to add the reference initial condition q0 as a field to ADSDomain? Then, an ADS integration can be a call to taylorinteg on an initial condition <: ADSDomain, in a similar way as JT is a dispatch on Taylor1{TaylorN{T}}.

I decided to concentrate all necessary parts of ADS in a single struct, then `taylorinteg` can use the ADS integrator via dispatch.

PerezHz:
Ok I see, thanks! I'm all in favor of condensing things as much as we can and avoid repeating ourselves, and I do see the value of keeping all ADS related things in one struct; I'm just wondering at the design level (or abstraction level, if you will) whether in this case it makes sense to actually put together everything in one place. What I mean is, for me the notion of an ADSDomain as an initial condition for an ADS integration, is different from an ADSTaylorSolution as the output of an ADS integration. So I would either try to keep ADSDomain with the minimal things we need, or keep the dispatch on `ADSTaylorSolution` well documented.

PR #186, opened 2024-03-26
Tests with Aqua.jl

lbenet:


lbenet:
Tests are now passing in nightly too

PerezHz:
Many thanks for adding this @lbenet! Is this PR still WIP, or is it ready for review? (actually changes look pretty much straight-forward so far)

lbenet:
It is essentially ready; I don't think there are too many additions to make. Also, I don't know if we should tag a new patch version...

test/aqua.jl
```diff
@@ -0,0 +1,36 @@
...
+@testset "Aqua tests (additional)" begin
+    Aqua.test_undefined_exports(TaylorIntegration)
+    Aqua.test_deps_compat(TaylorIntegration)
+    Aqua.test_stale_deps(TaylorIntegration;
+        ignore=[:DiffEqBase, :RecursiveArrayTools, :Requires, :StaticArrays])
+    Aqua.test_piracies(TaylorIntegration)
+    Aqua.test_unbound_args(TaylorIntegration)
+    Aqua.test_project_extras(TaylorIntegration)
+    Aqua.test_persistent_tasks(TaylorIntegration)
+end
```
PerezHz, line 35: Took a quick look to the Aqua docs, seems like there's a do-it-all function here, but I this is exactly equivalent to call the tests one by one as is it done now?
```suggestion
@testset "Aqua tests (additional)" begin
    Aqua.test_all(
    TaylorIntegration;
    stale_deps=(ignore=[:DiffEqBase, :RecursiveArrayTools, :Requires, :StaticArrays],),
    )
end
```

PerezHz (approved):
Just left a minor suggestion which I think is equivalent (though perhaps a bit more compact) to what you're already proposing. Either way, LGTM!

PerezHz:
> Also, I don't know if we should tag a new patch version...

I would actually agree to bump the patch version

test/aqua.jl
lbenet: Committed your suggestion; thanks!

lbenet:
Merging...

PR #189, opened 2024-04-06
Aqua: do not test ambiguities in Base and Core

PerezHz:
When called via `Aqua.test_all`, `Aqua.test_ambiguities` tests ambiguities in Base and Core together with TaylorIntegration, but that produces lots of ambiguities coming from other packages. Currently, #178 tests are failing due to this. Similar to what we did in NEOs, this PR changes Aqua tests to avoid detecting ambiguities in Base and Core (which in turn are produced mainly by ForwardDiff.Dual, StatsBase.PValue and StatsBase.TestStat among others), which are currently not used here.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TaylorIntegration"
 uuid = "92b13dbe-c966-51a2-8445-caca9f8a7d42"
 repo = "https://github.com/PerezHz/TaylorIntegration.jl.git"
-version = "0.15.1"
+version = "0.15.2"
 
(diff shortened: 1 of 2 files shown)


PerezHz:
(also triggered [this CI run](https://github.com/PerezHz/TaylorIntegration.jl/actions/runs/8580554157) to check how these changes look when #178 is applied on top of this PR)

PerezHz:
One check failed due to a weird coverage processing error but tests actually passed, so I'll go ahead and merge this.

PR #197, opened 2024-08-05
Update jetcoeffs.jl

miguelraz:
Pasan tests!


PerezHz:
Hola! 😄 

Makes a lot of sense to avoid the copy there, thanks for the fix! I'd only ask if you've tested if there are noticeable speedups from this change vs current `main`? If this is ready I'd only suggest to bump the patch version and then I think we can merge.

PerezHz:
LGTM so I'll go ahead and merge, thanks!

miguelraz:
Sorry, I didn't get a chance to benchmark vs `main`.

PR #203, opened 2024-12-09
Cache structs to save memory use on repeated calls to `taylorinteg!`

PerezHz:
This PR introduces some cache structs which allow reusing memory on repeated calls to `_taylorinteg!`.


PerezHz:
This is ready for review @lbenet @LuEdRaMo

lbenet:
Thanks @PerezHz for the nice addition. Do you have some comparisons/benchmarks? I'll have a look into it in the next few days!

LuEdRaMo:
@PerezHz: If I create a cache object once and then reuse it, do I have to update de `cache.t`, `cache.x` and `cache.dx` variables before every integration? If so, I think it would be nice to have a function for this.

PerezHz:
> @PerezHz: If I create a cache object once and then reuse it, do I have to update de cache.t, cache.x and cache.dx variables before every integration? If so, I think it would be nice to have a function for this.

Thank you for the suggestion! Right now, the workflow with these caches looks as follows:

```julia
# allocation
cache = TaylorIntegration.init_cache(Val(true), t0, q0, maxsteps, order)
# handle parsing
parse_eqs, rv = TaylorIntegration._determine_parsing!(true, f!, cache.t, cache.x, cache.dx, params);
# first integration
sol = taylorinteg!(Val(true), f!, q0, t0, tmax, abstol, rv, cache, params; parse_eqs, maxsteps)
# update
t0 = # new time
q0 = # new initial condition
# second integration
sol2 = taylorinteg!(Val(true), f!, q0, t0, tmax, abstol, rv, cache, params; parse_eqs, maxsteps)
```

That is, the cache is initialized only once, and the update to the new initial condition is handled internally by `taylorinteg!`.

PerezHz:
Tests are passing, so I'll go ahead and merge; if any fix is needed we can come back to this later

PR #205, opened 2025-05-14
Two minor changes

lbenet:
This PR proposes two minor changes:
- rename `update!` to `update_cache!`; TaylorSeries has a function called `update!`, so if both libraries are used one must qualify the module of the method used. I also think that the proposed name is more appropriate.
- In order to define the stepsize, we use the minimum, begining with a `convert(R,Inf)`. In this PR I propose to use `typemax(R)` (`R` is the proper type for the step size).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TaylorIntegration"
 uuid = "92b13dbe-c966-51a2-8445-caca9f8a7d42"
 repo = "https://github.com/PerezHz/TaylorIntegration.jl.git"
-version = "0.17.0"
+version = "0.17.1"
 
 [deps]
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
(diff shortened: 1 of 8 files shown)


lbenet:
(I'm upgrading TaylorModels to use the newest versions of TaylorSeries, TaylorIntegration and IntervalArithmetic, which is the motivation to propose these changes.)

lbenet:
@PerezHz Any idea of how to solve the tests in nightly?

lbenet:
Shall I go ahead and merge this (bumping a patch version), despite the problems in nightly? (I need this for TaylorModels.jl)

PerezHz:
> @PerezHz Any idea of how to solve the tests in nightly?

It seems the errors are coming from some DiffEq/SciML dependencies which fail to precompile and it cascades from there; so I'd say it's not strictly on us to fix

> Shall I go ahead and merge this (bumping a patch version), despite the problems in nightly? (I need this for TaylorModels.jl)

I agree to go ahead, perhaps at this point it's just better to just wait for the fix; if it turns out we need to patch things here we can do it afterwards.

lbenet:
Merging

PR #213, opened 2025-09-20
Fixes in common interface

lbenet:


lbenet:
Grrrr, I don't understand the problems with coveralls occurring sometimes in the tests...

PR #227, opened 2026-04-24
Migrate DiffEq extension for OrdinaryDiffEqCore+DiffEqBase sublibraries

PerezHz:


schillic:
Just my 2 cents: from experience, this change is not good because you lose the ability to react to breaking changes in DiffEqBase. The package manager will use the breaking version at a seemingly arbitrary point (when OrdinaryDiffEq.OrdinaryDiffEqCore decides to support the new version) and then your package may suddenly break.

PerezHz:
Thanks for the input @schillic! Yeah it looks like this transition is somehow not complete in OrdinaryDiffEq right? Maybe it would be better to just wait?

PerezHz:
This change was motivated by DiffEqBase being archived on April 21 this year:

https://github.com/SciML/DiffEqBase.jl

And has been moved as a sublibrary in OrdinaryDiffEq; so I understand it will be updated from there?

schillic:
I see. But if I understand correctly, this is purely a moving of the code, and the behavior should not depend on where the code is hosted. You can still install DiffEqBase without installing OrdinaryDiffEq, so my point still stands that DiffEqBase could suddenly make a breaking release without OrdinaryDiffEq making one and that could break things for you.

PerezHz:
As posted on [Slack](https://julialang.slack.com/archives/C7T968HRU/p1777214134954989) this change is ongoing so it might be better to wait a bit to adapt until things stabilize

PerezHz:
I think I managed to make the move to the newest releases of OrdinaryDiffEqCore, DiffEqBase, RecursiveArrayTools while keeping control on the compat constraint by including them explicitly in the compat section in Project.toml (like they are currently on `main`). Note this PR is dependent on https://github.com/JuliaDiff/TaylorSeries.jl/pull/407, and that will have to be merged first. Once that one is ready, this PR should be updated accordingly.

PerezHz:
Found some issues with the doc build, so trying to widen the compat bounds instead of excluding

PerezHz:
The widened compat bounds seem to be working; so as soon as the TaylorSeries update is ready I'll continue to merge this. As a bonus point, this PR manages to remove the full `OrdinaryDiffEq` as a weakdep and instead rely on `OrdinaryDiffEqCore`.

schillic:
Maybe rename this PR to reflect the new changes.

lbenet:
Thanks @PerezHz for the additions; https://github.com/JuliaDiff/TaylorSeries.jl/pull/407 has been merged and the patch version release is on its way.

PerezHz:
> Thanks @PerezHz for the additions; https://github.com/JuliaDiff/TaylorSeries.jl/pull/407 has been merged and the patch version release is on its way.

Many thanks @lbenet! I've updated this PR accordingly (including the title, as suggested by @schillic)  and tests pass locally, so CI should also pass, hopefully

PerezHz:
@lbenet after the last update I found a couple of issues with the docs build but have fixed those now. This is now ready for review

.github/workflows/ci.yml
```diff
@@ -59,7 +59,7 @@ jobs:
             ${{ runner.os }}-test-
             ${{ runner.os }}-
       ### Uncomment line below to test using a dependency branch
-      # - run: julia --project=. -e 'import Pkg; Pkg.add(url="https://github.com/PerezHz/TaylorSeries.jl", rev="jp/evaluate-tn"); Pkg.instantiate()'
+      # - run: julia --project=. -e 'import Pkg; Pkg.add(url="https://github.com/PerezHz/TaylorSeries.jl", rev="jp/fix-rat-compat"); Pkg.instantiate()'
```
lbenet: Is this needed? I guess not...

lbenet (commented):
I just left a comment which uses certain branch of TaylorSeries for the tests, which is already merged. Aside from that, LGTM.
Thanks a lot!

.github/workflows/ci.yml
PerezHz: Thanks for spotting this; the intention was in principle to have this commented in case we want to test TI using custom dependencies, but I agree can be misleading, so it's probably best to remove it (I already did).

PerezHz:
Thanks for the suggestion @lbenet! I fixed that and was expecting to get green lights from the CI to merge, but tests in nightly are failing... I already tried rerunning but the error is still there. I'd suggest to merge and then tackle that issue if it persists somehow, what would be your thoughts on this?

lbenet:
Before merging, what about #230 ? It seems to me very close to this PR. Maybe some `cherry-pick`ing would make both PRs more robust?

lbenet:
I actually like the minimalism of this PR 😄

lbenet:
Regarding the errors in nightly, seem to be related with some broadcasting... see e.g. [this](https://github.com/PerezHz/TaylorIntegration.jl/actions/runs/25077480536/job/73486337800?pr=227#step:6:325)

PerezHz:
For the fix for julia nightly see: https://github.com/JuliaLang/julia/issues/45086, and references therein

PerezHz:
Updated this PR after merging #230, and fixed errors in nightly, so will try to merge as soon as CI gives the green lights

lbenet:
Thanks a lot!

PR #240, opened 2026-06-03
Add (internal) `@taylorize2`

lbenet:
Currently, `@taylorize` adds two internal methods (`jetcoeffs!` and `_allocate_jetcoeffs!`) that allow to improve performance and allocations when integrating with `parse_eqs=true` (which is the current default). It turns out that it is easy to modify it to produce a new method of the RHS of the equations of motion (i.e., `f(u, p, t)` or `f(du, u , p, t)`), which is *taylorized* in the same sense. (The signature of the new method is `f(u, p, t, rv::RetAlloc)`, or `f(du, u, p, t, rv::RetAlloc)`.) Yet, the new method sometimes creates issues that break the interface with `OrdinaryDiffEq`. (I think I understand the reasons for this, and a work around it, but it is *unconfortable* from the user perspective.) This is the reason that I propose a new (internal) macro `@taylorize2`. That method finds applications in TaylorModels, to improve performance, though I find *natural* that the code remains here, which makes maintaning it simpler.


test/taylorize.jl
```diff
@@ -480,9 +545,12 @@ import Logging: Warn
     # Complex dependent variables
     @testset "Complex dependent variable" begin
         cc = complex(0.0, 1.0)
-        @taylorize eqscmplx(x, p, t) = cc * x
+        TaylorIntegration.@taylorize2 eqscmplx(x, p, t) = cc * x
 
         @test (@isdefined eqscmplx)
+        @test length(methods(eqscmplx)) == 2
+
+        # Mie 10, 5pm; 17 (10:30 o por la tarde); 24 (10:30 o por la tarde)
```
PerezHz: I think this has to be removed
```suggestion
```

PerezHz (approved):
LGTM! This looks really useful as an internal helper (and in the future, for many other things), many thanks @lbenet!

PR #243, opened 2026-06-07
Miscellaneous performance improvements

PerezHz:
This PR aims at improving overall performance of the package from mostly two different angles:

- On JT integrations, `stepsize` noticeably allocates. Here, we fix this by "taylorizing" by hand this method.
- Avoid `deepcopy` and instead rely on`TaylorSeries.identity!`: this avoids extra work by `deepcopy` not needed here, while maintaning full independency of objects in memory where needed (i.e., avoiding aliasing) and minimally allocating as much as possible.
- Other scattered fixes.

### DIFF
--- a/src/integrator/taylorinteg.jl
+++ b/src/integrator/taylorinteg.jl
@@ -1,5 +1,32 @@
 # This file is part of the TaylorIntegration.jl package; MIT licensed
 
+# _store_taylor!
+@inline function _stored_taylor(x::Taylor1{U}) where {U<:Number}
+    y = zero(x)
+    TS.identity!(y, x)
+    return y
+end
+
+@inline function _store_taylor!(psol::Array{Taylor1{U},1}, nsteps::Int,
+        x::Taylor1{U}) where {U<:Number}
+    if isassigned(psol, nsteps)
+        @inbounds TS.identity!(psol[nsteps], x)
+    else
+        @inbounds psol[nsteps] = _stored_taylor(x)
+    end
+    return nothing
+end
+
+@inline function _store_taylor!(psol::Array{Taylor1{U},2}, j::Int, nsteps::Int,
+        x::Taylor1{U}) where {U<:Number}
+    if isassigned(psol, j, nsteps)
+        @inbounds TS.identity!(psol[j, nsteps], x)
+    else
+        @inbounds psol[j, nsteps] = _stored_taylor(x)
+    end
+    return nothing
+end
+
 # set_psol!
 @doc doc"""
     set_psol!(::Val{true}, psol::Array{Taylor1{U},1}, nsteps::Int, x::Taylor1{U}) where {U<:Number}
@@ -19,7 +46,7 @@ field `:p` in [`TaylorSolution`](@ref). See also [`init_psol`](@ref).
     nsteps::Int,
     x::Taylor1{U},
 ) where {U<:Number}
-    @inbounds psol[nsteps] = deepcopy(x)
+    _store_taylor!(psol, nsteps, x)
     return nothing
 end
 @inline function set_psol!(
@@ -28,7 +55,9 @@ end
     nsteps::Int,
     x::Vector{Taylor1{U}},
 ) where {U<:Number}
-    @inbounds psol[:, nsteps] .= deepcopy.(x)
+    @inbounds for j in eachindex(x)
+        _store_taylor!(psol, j, nsteps, x[j])
+    end
     return nothing
 end
 @inline set_psol!(::Val{false}, args...) = nothing
--- a/src/integrator/taylorinteg.jl
+++ b/src/integrator/taylorinteg.jl
@@ -40,7 +40,18 @@ Matching `Taylor1` and `TaylorN` values with compatible lengths are updated in
 place with `TaylorSeries.identity!`, preserving the object already held by the
 cache. If the Taylor polynomial storage is incompatible, or if the value is an
 immutable scalar-like number, the method returns a fresh [`_stored_value`](@ref)
-instead. Callers must assign the return value back into the storage slot.
+instead.
+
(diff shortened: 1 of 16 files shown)


src/integrator/taylorinteg.jl
```diff
@@ -1,5 +1,133 @@
...
+Copy `src` into reusable storage `dest` when possible and return the stored
+value. Matching `Taylor1` and `TaylorN` values are updated with
+`TaylorSeries.identity!`; incompatible or immutable values are replaced by a
+fresh stored value.
+"""
+@inline function _copy_value!(dest::Taylor1{U}, src::Taylor1{U}) where {U<:Number}
+    length(dest) == length(src) || return _stored_value(src)
+    TS.identity!(dest, src)
+    return dest
+end
```
lbenet, line 50: I don't think I understand this function. 

If the lengths are the same, which I would think is the common case, `_stored_value` is some sort of alias to `_stored_taylor`, which at the end of the day returns a new object(!) using `identity!`, which is identical to `src`. I don't see that `dest` is updated in this case. Moreover, if the lengths are not identical, `identity!` is also called. So, why not simply use `identity!` directly...

src/rootfinding.jl
```diff
@@ -1,12 +1,27 @@
...
     build_solution(t, x, tevents, xevents, gresids, nsteps, nevents)
+    build_solution(copy_solution::Val, t, x, p, tevents, xevents, gresids, nsteps, nevents)
+    build_solution(copy_solution::Val, t, x, tevents, xevents, gresids, nevents)
 
 Helper function to build a [`TaylorSolution`](@ref) from a call to a
 root-finding method of [`taylorinteg`](@ref).
 
+When `copy_solution` is `Val(false)`, the returned solution borrows views of the
+given arrays. When `copy_solution` is `Val(true)`, `t`, `x`, `p`, `tevents`,
+`xevents` and `gresids` are copied into independent owned storage.
```
lbenet, line 27: I would guess that the `view`s are more performant, so what's an use case for the `Val(true)` case?

test/many_ode.jl
```diff
@@ -37,6 +37,41 @@ import Logging: Warn
...
+            )
+        end
+
+        function norm_based_stepsize(q::AbstractVector, epsilon)
+            h = typemax(Float64)
+            for i in eachindex(q)
+                h = min(h, norm_based_stepsize(q[i], epsilon))
+            end
+            return h
+        end
```
lbenet: I think these functions are equivalent to `stepsize(q, absepsilon, relepsilon)`, so why not using those? I see that you are testing below `stepsize` so you compare the output with the results of `norm_based_stepsize`, but isn't it a trivial check?

PerezHz:
@lbenet thank you for your comments; to help clarify how https://github.com/JuliaDiff/TaylorSeries.jl/pull/415 affects TaylorIntegration I have opened https://github.com/PerezHz/TaylorIntegration.jl/pull/245, which contains the minimal changes in TI needed because of https://github.com/JuliaDiff/TaylorSeries.jl/pull/415. I propose to merge those other two PRs, and then we can come back to this one, which includes further changes. Sorry for the back and forth.

src/rootfinding.jl
PerezHz: This is mainly related to object ownership: `taylorinteg!` (with the bang `!`) allows users to reuse caches, i.e., it uses a shared memory model, but since the `.psol` field is included in the cache, when that cache is reused to construct another solution, `psol` of the previous solution is overwritten with the new solution. So in this case we want `copy_solution = Val(true)`, which is essentially saying reuse caches, but construct independent solution objects which "own" all their fields (thus, an independent copy is made at solution object construction).

The case `copy_solution = Val(false)` means users will not necessarily reuse the cache (e.g., by using `taylorinteg` (non-banged `!` version), and thus `psol` in the cache can be shared with the corresponding field in the solution object, and in that case it's safe to alias (i.e., not allocate).

There's probably other approaches to this problem, but this was a good compromise I found between not using `deepcopy` for everything, while being able to reuse caches for propagation, (in `NEOs.jl` caches are reused a lot, and for good reason).

PerezHz: See e.g. [here](https://github.com/PerezHz/NEOs.jl/blob/856fae099416a8b416bee2ca172c237415003206/src/propagation/propagation.jl#L106)

src/integrator/taylorinteg.jl
PerezHz: You're right that `_stored_taylor` was essentially an alias for `_stored_value`; I have removed the former. The difference is precisely the allocation of a new object; I'm trying to find an easier way to do this. The problem is, even two in the method body the values of `dest` and `src` are known, the place (slot) in memory where `dest` will be saved is not known, so maybe this mechanism should be changed so that the slot is passed and updated accordingly.

PerezHz: I reorganized this part of the code a bit, but left `_stored_value` and `_copy_value!` for now since still represent two genuinely different operations: `_stored_value(src)` means "an independent stored value is needed, and there may not be a destination yet"; `_copy_value!(dest, src)` means "there's already a available destination in memory; reuse it if compatible". So yes, both paths use `identity!`, but one copies into fresh storage and the other copies into existing storage. I would argue that distinction is important enough to keep. The direct way to use this functionality is now `_store_value!`, which hides that choice:
```julia
_store_value!(dest_array, src, i)
_store_value!(dest_matrix, src, i, j)
```

test/many_ode.jl
PerezHz: Thanks; removed these duplicated functions and switched to a direct comparison.

PerezHz:
I'll go ahead with the merge, and we can certainly revisit this in case any issues arise.

